PR #1, opened 2022-01-13
WIP: Improve installer

rouson:
Work In Progress: **Do Not Merge**
=========================
This pull request updates the installer to be more clear about explaining the groupings of prerequisites that it installs (e.g., either all compilers or none) and to ask permission before each prerequisite group gets installed.

PR #2, opened 2022-01-14
Disaggregate prerequisites

rouson:
This PR replaces #1 and addresses several [PR comments] and [issue comments] made on a predecessor fork about the Caffeine installer.  The PR enhances the installer to

1. Replace `#!/bin/dash` with the more portable `#!/bin/sh`.
2. Always ask for permission before installing a package,
3. Produce output that explains which packages will be installed if permission is granted,
4. Ask permission to install and use Homebrew for package installation if any Homebrew-installable packages are missing, and
5. Disaggregate prerequisites where possible\*.

On my machine, this PR completes without error 
- [X] On a bare macOS system without Homebrew pre-installed and with no prerequisites pre-installed.  
- [X] On macOS with all prerequisites installed.
- [X] Inside a Lubuntu Linux virtual machine without Homebrew pre-installed and with no prerequisites pre-installed.
- [X] Inside a Lubuntu Linux virtual machine with all prerequisites installed.

\* Because some Homebrew packages contain more than one prerequisite that cannot be installed independently of each other, there will be some cases in which an already installed prerequisite will nonetheless be installed via Homebrew.  For example, if any of the environment variables `FC`, `CC`, and `CXX` is unset and any of the Homebrew `gcc-11`, `g++-11`,  and `gfortran-11` is not in the `PATH`, then the script will excecute `brew install gcc@11` if granted permission and this will result in the installation of all three compilers:  `gcc-11`, `g++-11`,  and `gfortran-11`.

[PR comments]: https://github.com/rouson/caffeine/pull/15
[issue comments]: https://github.com/rouson/caffeine/issues/14

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -169,8 +169,9 @@ if [ -z ${FC+x} ] || [ -z ${CC+x} ] || [ -z ${CXX+x} ] || [ -z ${PKG_CONFIG+x} ]
   CC=`which gcc-$GCC_VER`
   CXX=`which g++-$GCC_VER`
   FC=`which gfortran-$GCC_VER`
-  PREFIX=`realpath $PREFIX`
 fi
+
+PREFIX=`realpath $PREFIX`
 
 export FPM_FC="$FC"
 export FPM_CC="$CC"
--- a/install.sh
+++ b/install.sh
@@ -14,6 +14,8 @@ print_usage_info()
     echo " --prereqs          Display a list of prerequisite software."
     echo "                    Default prefix='\$HOME/.local/bin'"
     echo ""
+    echo "For a non-interactive build with the 'yes' utility installed, execute"
+    echo "yes | ./install.sh"
 }
 
 GCC_VERSION=11
--- a/install.sh
+++ b/install.sh
@@ -61,12 +61,6 @@ done
 
 set -u # error on use of undefined variable
 
-PREFIX=${PREFIX:-"$HOME/.local"}
-echo "Using installation prefix $PREFIX"
-
-PKG_CONFIG_PATH=${PKG_CONFIG_PATH:-"$PREFIX/lib/pkgconfig"}
-echo "Using pkg-config prefix $PKG_CONFIG_PATH"
-
 if [ -z ${FC+x} ] || [ -z ${CC+x} ] || [ -z ${CXX+x} ]; then
   if command -v gfortran-$GCC_VER > /dev/null 2>&1; then
     FC=`which gfortran-$GCC_VER`
@@ -125,6 +119,12 @@ ask_permission_to_install_homebrew_package()
   fi
   printf "Is it ok to use Homebrew to install $1? [yes] "
 }
+
+PREFIX=${PREFIX:-"$HOME/.local"}
+echo "Using installation prefix $PREFIX"
+
+export PKG_CONFIG_PATH=${PKG_CONFIG_PATH:-"$PREFIX/lib/pkgconfig"}
+echo "Using pkg-config prefix $PKG_CONFIG_PATH"
 
 exit_if_user_declines()
 {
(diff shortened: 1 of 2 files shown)


rouson:
@bonachea yes, this PR is mean to replace PR #1, which I'll close.  Thanks for the feedback. I'll switch to your suggested curl command.

bonachea:
> this is from command: `curl -L $GASNET_SOURCE_URL | tar xvz - -C $DEPENDENCIES_DIR`
> 
> This system has "tar (GNU tar) 1.26", and I think the command usage above may be newer/less portable (specifically, omitting the `f`). FWIW UPC++ uses this command: `gzip -cd "$GASNET" | ( cd bld && tar xf - )` which seems to work pretty universally.

The following change was sufficient to make it work (adding the missing `f`):

```
--- a/install.sh
+++ b/install.sh
@@ -217,7 +217,7 @@ if [ ! -f "$PKG_CONFIG_PATH/$pkg.pc" ]; then
     mkdir -pv $DEPENDENCIES_DIR
   fi

-  curl -L $GASNET_SOURCE_URL | tar xvz - -C $DEPENDENCIES_DIR
+  curl -L $GASNET_SOURCE_URL | tar xvzf - -C $DEPENDENCIES_DIR

   if [ -d $DEPENDENCIES_DIR/GASNet-$GASNET_VERSION ]; then
     cd $DEPENDENCIES_DIR/GASNet-$GASNET_VERSION
~                                                   
```

bonachea:
Later in the install I got to this failure:

```
make[2]: Leaving directory '/home/pcp1/bonachea/UPC/caffeine/build/dependencies/GASNet-2021.9.0'
make[1]: Leaving directory '/home/pcp1/bonachea/UPC/caffeine/build/dependencies/GASNet-2021.9.0'
/home/pcp1/bonachea/UPC/caffeine
/home/pcp1/bonachea/UPC/caffeine
/home/pcp1/bonachea/UPC/caffeine
./build/run-fpm.sh: line 3: /home/pcp1/bonachea/.local/bin/fpm: No such file or directory
```

Note this is using a /usr/local install of `fpm` I placed in my `$PATH`, NOT one installed by install.sh (which the file generation for `run-fpm.sh` appears to be assuming)

install.sh
```diff
@@ -38,63 +61,177 @@ while [ "$1" != "" ]; do
...
+    mkdir -pv $DEPENDENCIES_DIR
+  fi
+  
+  curl -L $GASNET_SOURCE_URL | tar xvz - -C $DEPENDENCIES_DIR
+  
+  if [ -d $DEPENDENCIES_DIR/GASNet-$GASNET_VERSION ]; then
+    cd $DEPENDENCIES_DIR/GASNet-$GASNET_VERSION
+      FC="$FC" CC="$CC" CXX="$CXX" ../GASNet-$GASNET_VERSION/configure --prefix "$PREFIX"
       make -j 8 all
       make check
```
bonachea, line 230: GASNet `make check` builds a bunch of GASNet tests (which ensures they link), but (with a few trivial exceptions) does not run them.  Running multi-node GASNet tests often requires a working spawning setup or even a batch environment, so separate command `make run-tests` is used to actually run the tests.

The GASNet test building step takes a significant amount of disk space and time (especially with serial make), and running the tests would take even longer. This might make sense inside a caffeine self-test/CI, but really doesn't make sense for an installer. Given the caffeine build system just wants to use GASNet (not test it), I'd suggest deleting this `make check` step. 

On a related note, the `make install` step below can also benefit from parallel make eg `-j 8`.

rouson: `make check` has been dropped in the latest commit.

rouson:
@everythingfunctional it appears the macOS CI is failing on a [linking error]. I wonder if there's an issue with the flags in the generated `fpm.toml` manifest.  Could you investigate this?

[linking error]: https://github.com/BerkeleyLab/caffeine/runs/4815223140?check_suite_focus=true#step:4:2219

rouson:
At last, the Linux CI tests pass!  It seems that the macOS CI tests have been hanging on the last few pushes, which appears to be a GitHub issue so I'll likely merge this PR if the GitHub CI issue doesn't resolve itself shortly.

rouson:
The macOS logs show the installer running to completion without any errors, but the process continues until being killed after 360 minutes. I'll merge this.

PR #5, opened 2022-01-19
Partial co_sum support: integer(c_int32) A argument of rank 0 or 1 and  no optional arguments 

rouson:
This PR adds support for the `co_sum` collective subroutine, covering the use case of no optional arguments and a `c_int32` argument `A` of rank 0 or 1.


rouson:
Something funky is going on with the macOS CI tests.   The CI completes without error, but the process hangs anyway.   I'm going to disable the macOS CI testing for now because it's convenient for me to test on macOS locally.  The Linux CI testing will remain.

PR #7, opened 2022-01-21
Add co_sum support for all valid types & ranks

rouson:
With this PR, `caf_co_sum` supports all possible argument types and ranks of the following `kind`:
- [X] default `integer` and `integer(c_int64_t)` arguments of all possible ranks (0-15)
- [X] default `real` and `double precision` arguments of all possible ranks (0-15)
- [X] default and `double precision` `complex` arguments of all possible ranks (0-15)

### DIFF
--- a/src/caffeine/collective_subroutines_s.f90
+++ b/src/caffeine/collective_subroutines_s.f90
@@ -66,6 +66,10 @@ contains
            call c_co_sum_no_result_image_float(c_loc(a), nelem=size(a,kind=c_size_t))
          type is(real(c_double))
            call c_co_sum_no_result_image_double(c_loc(a), nelem=size(a,kind=c_size_t))
+         type is(complex(c_float))
+           call c_co_sum_no_result_image_float(c_loc(a), nelem=size(a,kind=c_size_t))
+         type is(complex(c_double))
+           call c_co_sum_no_result_image_double(c_loc(a), nelem=size(a,kind=c_size_t))
          class default
            error stop "caf_co_sum: unsupported type"
        end select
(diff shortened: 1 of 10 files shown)


src/caffeine/collective_subroutines_s.f90
```diff
@@ -10,30 +11,306 @@
...
+         type is(integer(c_int64_t))
+           call c_co_sum_no_result_image_int64(c_loc(a), nelem=1_c_size_t)
+         type is(real(c_float))
+           call c_co_sum_no_result_image_float(c_loc(a), nelem=1_c_size_t)
+         type is(real(c_double))
+           call c_co_sum_no_result_image_double(c_loc(a), nelem=1_c_size_t)
+         type is(complex(c_float))
+           call c_co_sum_no_result_image_float(c_loc(a), nelem=2_c_size_t)
+         type is(complex(c_double))
+           call c_co_sum_no_result_image_double(c_loc(a), nelem=2_c_size_t)
```
bonachea, line 55: Forgive my ignorance, where are 1_c_size_t and 2_c_size_t defined ?

bonachea:
Where do you inspect/consume the optional `result_image` argument?

It occurs to me that rather than double your combinatorially exploding code size yet again, perhaps you should just arrange to pass result_image to the C file (eg using `result_image=-1` to mean "all"), and let it branch on whether to call GASNet reducetoone or reducetoall?

src/caffeine/collective_subroutines_s.f90
rouson: These are what Fortran calls literal constants so they are just numbers.   The  `_c_size_t` suffix tells the Fortran compiler that I want the preceding number to be stored in a bit representation that exactly matches the companion C compiler's `size_t` representation.  According to the Fortran standard, every Fortran compiler must have a companion C compiler.  In most cases, it's the C compiler from the same vendor with the exception being vendors that don't produce a C compiler and therefore designate a third-party C compiler as the companion C compiler.

rouson:
@bonachea I should also mention that there are actually multiple optional arguments in Fortran's collective subroutines.  I'll replace mentions of `result_image` with just "optional arguments."

PR #10, opened 2022-01-27
Add optional stat arguments to co_sum, co_min, & co_max

rouson:


ktras:
@rouson In `caf_co_max_test.f90` you did not assign the variable `status` a value before passing it to the collective, whereas in the other tests, you did so. Did you mean for `caf_co_max_test.f90` to differ like that? If that was your intent, then this looks good to me.

rouson:
@ktras great catch.  I meant to initialize it to ensure that it gets redefined to `0` upon exiting each invoked C function.

rouson:
I just pushed a fix and will merge after tests pass.

PR #11, opened 2022-01-29
Feature: support `result_image` & simplify C reduction code

rouson:
This PR 
1. Adds the functionality that supports the optional `result_image` argument  in
- [X] `co_min`
- [X] `co_max`
- [X] `co_sum`
2. Simplifies `caffeine.c` by adding a `c_co_reduce_universal` function that each of the other `c_co_*` functions each invokes, reducing the latter functions to one executable line.


src/caffeine/caffeine.c
```diff
@@ -53,110 +53,77 @@ void c_sync_all()
   gasnet_barrier_notify(0,GASNET_BARRIERFLAG_ANONYMOUS);
   gasnet_barrier_wait(0,GASNET_BARRIERFLAG_ANONYMOUS);
 }
-void c_co_sum_no_result_image_int32(void* c_loc_a, size_t Nelem, int* stat)
+
+void c_co_reduce_universal(void* c_loc_a, size_t Nelem, int* stat, int* result_image, int gex_dt, size_t sizeof_type, int gex_op)
```
bonachea, line 57: As a general rule I'd advise you not to define identifiers with a `gex_` or `gasnet_` name prefix. 

These name prefixes "belong" to GASNet, so it's not impossible that someday the GASNet headers might add a macro or function symbol aliasing the variable names you've chosen above. This could lead to compile errors in files including the GASNet headers, or even link-time symbol conflicts if both are extern symbols.

If you're looking for something that "says GASNet" but is not reserved, perhaps use `gn_`?

bonachea, line 57: Slapping an `inline` keyword on `c_co_reduce_universal` is probably a good idea, although a good enough optimizer might figure that out.

If you want a portable demand for inlining, you can instead use `GASNETT_INLINE(c_co_reduce_universal)` immediately before the function prototype ([docs](https://gasnet.lbl.gov/dist-ex/README-tools))

The motivation is that inlining allows the GASNet call to "see" the manifest constant parameters provided in the callers; I just checked and this doesn't actually make much difference for the current Reduce implementation, but it makes a big difference for some other GASNet communication calls (notably RMA and atomics).

src/caffeine/collective_subroutines/co_max_s.f90
```diff
@@ -11,61 +11,67 @@
...
      stat_ptr = c_loc(stat)
    else
      stat_ptr = c_null_ptr
    end if
 
+   if (present(result_image)) then
+     result_image_ptr = c_loc(result_image)
+   else
+     result_image_ptr = c_null_ptr
+   end if
```
bonachea, line 59: Why the level of indirection on `result_image`?

IIUC this forces the Fortran compiler to spill the result_image value to the stack in order to create a valid pointer, where it otherwise could remain in a register.

IIUC you could instead arrange to pass an `int` value argument to GASNet, where `0` means "all" (since Fortran images are 1-based).

bonachea (changes requested):
Overall looks like a great improvement!

I added some suggestions that could make it even stronger.

src/caffeine/caffeine.c
rouson: Good suggestion.  I think I'll just drop the prefixes.

rouson: @bonachea Sounds good.  I'll try `GASNETT_INLINE`.  Out of curiosity, why the double `T?

bonachea: GASNETT_ is the namespace for the "GASNet Tools" interface, a collection of portability tools for running on different platforms/compilers/etc that are not directly relevant to communication, but provided as a service to our clients.

GASNet tools has its own spec here: https://gasnet.lbl.gov/dist-ex/README-tools

src/caffeine/collective_subroutines/co_max_s.f90
rouson: @bonachea   Wow. Really nice catch!  This is probably an example of copypasta (which isn't quite the write term but I loosely define it as spaghetti code that results from my tendency copy and paste working code at the beginning for writing new code).  I suspect that the only reason `result_image` is passed as a pointer is because it comes right after I `stat` in the argument list.  I think the arguments are listed in the order I implemented them so I probably copied whatever I was doing with `stat` and did the same with `result_image`!  I'll switch to just passing an `int` by value as you suggest.

rouson: @bonachea I remember why I chose to make it a pointer.  There are several subtleties that I had to investigate in the standard before I could be sure that it would be safe to choose a special value indicate that no `result_image` was specified by the user.  For example, if a coarray `A` has multiple codimensions, then `this_image(A)` returns a vector indicating the image's position in a multidimensional index space wherein the range of each vector component can span positive and negative values.  Your comment prompted me to research this issue in the standard and it appears that image numbers are required to be positive so it should work to use a zero or negative `result_image` value to as a signal that no `result_image` was specified.

src/caffeine/caffeine.c
rouson: @bonachea commit a1951b4 dropped the prefixes as suggested.

src/caffeine/collective_subroutines/co_max_s.f90
rouson: @bonachea commit 343ebb3 switches to passing `result_image` by value.

src/caffeine/caffeine.c
rouson: @bonachea I tried inserting `GASNETT_INLINE(c_co_reduce_universal)` on the line before `void c_co_reduce_universal(...` and got the compiler warning and errors below.  Because you mention "function prototype," I wonder if the problem is that I don't have a header file. I also thought the issue might be that the file isn't getting preprocessed so I tried adding `--cpp` in the `--c-flag` argument in `build/run-fpm.sh`.  

```
././src/caffeine/caffeine.c:57:1: warning: return type defaults to 'int' [-Wimplicit-int]
   57 | GASNETT_INLINE(c_co_reduce_universal)
      | ^~~~~~~~~~~~~~
././src/caffeine/caffeine.c: In function 'GASNETT_INLINE':
././src/caffeine/caffeine.c:59:1: error: expected '=', ',', ';', 'asm' or '__attribute__' before '{' token
   59 | {
      | ^
././src/caffeine/caffeine.c:74:1: error: expected '=', ',', ';', 'asm' or '__attribute__' before '{' token
   74 | {
      | ^
```
and more errors similar to the above ones.

bonachea: > @bonachea I tried inserting `GASNETT_INLINE(c_co_reduce_universal)` on the line before `void c_co_reduce_universal(...` and got the compiler warning and errors below. Because you mention "function prototype," I wonder if the problem is that I don't have a header file. I also thought the issue might be that the file isn't getting preprocessed so I tried adding `--cpp` in the `--c-flag` argument in `build/run-fpm.sh`.

I'm not sure what might be wrong without seeing the code you actually compiled - are you 100% certain you did not delete the `void` return type? The macro goes _before_ the return type, something like:

```
GASNETT_INLINE(c_co_reduce_universal)
void c_co_reduce_universal(void* c_loc_a, size_t Nelem, int* stat, int result_image, int dt, size_t sizeof_type, int op, gex_Coll_ReduceFn_t* user_op, void* client_data)
{
... function body...
```

It doesn't matter whether you have a header file or not, this should work regardless.

bonachea: Note that you ALSO need to `#include <gasnet_tools.h>` at the top - that's probably why it failed to compile..

rouson: added in 34e3cbc95de2d7971c3f698a58bfead5e573e51c

PR #12, opened 2022-01-29
Feature: support co_reduce

rouson:
As long as the tests pass, this PR can be merged at any time.  In the interim, the current partial support for `co_reduce` will be expanded through additional commits.

To Do
------
Provide  support for `operation` functions that accept operands of intrinsic type*:
- [X] default `integer`
- [ ]  `integer(c_int64_t)`
- [ ] default `real`
- [ ] `real(c_double)`
- [ ] default `complex
- [ ] `complex(c_double)`
- [ ] `character(c_char)`
- [ ] `logical(c_bool)` 

Future Work
-------------
Support user-defined derived type operands (objects) in a future PR.

\* The Caffeine procedures use types with C-interoperable kind values.  A subset of the unit tests use default types.  A test failure will result if the Fortran compiler's default type has a `kind` parameter that does not match any of the `kind` parameters for a  corresponding variable in Caffeine.  This approach provides a diagnostic to flag any need for broader `kind` support in Caffeine.


bonachea (changes requested):
I added some comments on pull request #11 which this PR appears to subsume, but I believe all my suggestions for improvement are still relevant to the code here.

rouson:
@bonachea correct: this PR subsumes #11. I made the requested changes in #11 and then got into a weird git state when I merged those changes into the branch in this PR so I have created #13 to replace this one.  I'll close this one.

rouson:
@bonachea  I also separately made the same suggested changes on the branch in #13.

PR #13, opened 2022-01-30
Replace PR #12: support co_reduce

rouson:
The text below was copied and pasted from #12, which this PR replaces.

As long as the tests pass, this PR can be merged at any time.  In the interim, the current partial support for `co_reduce` will be expanded through additional commits.

To Do
------
Provide  support for `operation` functions that accept operands of intrinsic type*:
- [X] default `integer`
- [ ]  `integer(c_int64_t)`
- [x] default `real`
- [ ] `real(c_double)`
- [ ] default `complex`
- [ ] `complex(c_double)`
- [x] `character(c_char)`
- [x] `logical(c_bool)` 

Future Work
-------------
Support user-defined derived type operands (objects) in a future PR.

\* The Caffeine procedures use types with C-interoperable `kind` values.  A subset of the unit tests use default types.  A test failure will result if the Fortran compiler's default type has a `kind` parameter that does not match any of the `kind` parameters for a  corresponding variable in Caffeine.  This approach provides a diagnostic to flag any need for broader `kind` support in Caffeine.


src/caffeine/collective_subroutines/co_reduce_s.f90
```diff
@@ -1,9 +1,227 @@
...
+          call c_co_reduce_char( &
+            c_loc(a), len_a, stat_ptr, optional_value(result_image), c_funloc(Coll_ReduceSub_c_char), c_loc(len_a))
+        end block
+      rank default
+         error stop "caf_co_reduce_c_char: unsupported rank"
+    end select
+
+  contains
+
+    subroutine Coll_ReduceSub_c_char(arg1, arg2_and_out, count, cdata) bind(C)
```
bonachea, line 54: I don't understand how this callback works, but would love to have you explain it to me in our next discussion

src/caffeine/collective_subroutines/co_reduce_s.f90
```diff
@@ -1,9 +1,227 @@
...
+
+    c_char_op_ptr => operation
+
+    stat_ptr = get_c_ptr(stat)
+
+    select rank(a)
+      rank(0)
+        block
+          integer(c_size_t), target :: len_a
+          len_a = int(len(a), c_size_t)
```
bonachea, line 44: IIUC, the requirement for A to have "the same shape, type, and type parameter values" means the character strings are guaranteed to be the same length on all images, right?

src/caffeine/caffeine.c
```diff
@@ -53,110 +54,98 @@ void c_sync_all()
...
+     c_co_reduce_universal(c_loc_a, Nelem, stat, result_image, GEX_DT_FLT, sizeof(float), GEX_OP_USER, operation, NULL);
 }
 
-void c_co_max_no_result_image_double(void* c_loc_a, size_t Nelem, int* stat)
+void c_co_reduce_char(void* c_loc_a, size_t Nelem, int* stat, int result_image, gex_Coll_ReduceFn_t* operation, void* client_data)
 {
-      gex_Event_t ev
-        = gex_Coll_ReduceToAllNB(myteam, c_loc_a, c_loc_a, GEX_DT_DBL, sizeof(double), Nelem, GEX_OP_MAX, NULL, NULL, 0);
-      gex_Event_Wait(ev);  
+     c_co_reduce_universal(c_loc_a, Nelem, stat, result_image, GEX_DT_USER, sizeof(char), GEX_OP_USER, operation, client_data);
```
bonachea, line 145: I think this is a bug. 

IIUC this is a user-defined operator reduction where each rank passes a character string with fixed length L, and as written ends up calling `gex_Coll_ReduceTo{One,All}NB(GEX_DT_USER,GEX_OP_USER)` with dt_sz==sizeof(char) and dt_cnt=L, but this is **probably wrong for character strings**. Specifically, we are not reducing an array of L independent 1-byte elements that GASNet is free to fragment and invoke the callback on smaller pieces (which I believe it WILL do, for sufficient large dt_cnt). Rather, this is a reduction over an L-byte datatype where each string is an indivisable datum as far as callbacks are concerned. So dt_sz  should be L, and dt_cnt should be 1 (unless/until you add support for an array of strings).

CC: @phhargrove

bonachea (changes requested):
I added a few comments, and pointed out several defects that need to be addressed

src/caffeine/caffeine.c
```diff
@@ -53,110 +54,98 @@ void c_sync_all()
...
+     c_co_reduce_universal(c_loc_a, Nelem, stat, result_image, GEX_DT_DBL, sizeof(double), GEX_OP_MAX, NULL, NULL);
 }
 
-void c_co_max_no_result_image_float(void* c_loc_a, size_t Nelem, int* stat)
+void c_co_reduce_int32(void* c_loc_a, size_t Nelem, int* stat, int result_image, gex_Coll_ReduceFn_t* operation)
 {
-      gex_Event_t ev
-        = gex_Coll_ReduceToAllNB(myteam, c_loc_a, c_loc_a, GEX_DT_FLT, sizeof(float), Nelem, GEX_OP_MAX, NULL, NULL, 0);
-      gex_Event_Wait(ev);  
+     c_co_reduce_universal(c_loc_a, Nelem, stat, result_image, GEX_DT_I32, sizeof(int32_t), GEX_OP_USER, operation, NULL);
```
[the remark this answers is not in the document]
rouson: This will be captured in release notes.  See #17

src/caffeine/caffeine.c
rouson: This comment is now linked in issue #18

rouson: @bonachea I have verified that this issue has been resolved in PR #21.  As of the current head of the corresponding branch, the `co_min`, `co_max`, and `co_reduce` tests employ a range of string lengths and string array shapes.  In each case, the call to `gex_Coll_ReduceTo{One,All}` results in `dt_sz` representing the the number of bytes in a  single string and `dt_cnt` being the number of strings that each rank holds in its string array (with `dt_cnt` being 1 when each rank just holds scalar, i.e., one `dt_sz`-byte string).  All tests pass, including tests with scalars, 1D, 2D, and 3D arrays, including one test in which the array size and element length are both set at runtime.  I'll close issue #18.

PR #15, opened 2022-02-07
Assumed-type first argument in `caf_co_sum`

rouson:
This PR greatly simplifies `caf_co_sum` by replacing the unlimited polymorphic (`class(*)`) first argument with an assumed-type (`type(*)`) argument and thereby eliminating all `select rank` and `select type` blocks.


rouson:
@bonachea I'm hoping you can help me diagnose odd behavior with one test, wherein the collective sum yields the correct answer on each image, but then the code crashes.  I can demonstrate during an upcoming GASNet-EX call or whenever is best for you.  The test is currently commented out at [line 23] in `test/caf_co_sum_test.f90`.

[line 23]: https://github.com/BerkeleyLab/caffeine/blob/905998c653ced6d9b20d0318e3bc3537c586e221/test/caf_co_sum_test.f90#L23

src/caffeine/caffeine.c
```diff
@@ -157,9 +137,61 @@ void c_co_broadcast(CFI_cdesc_t * a_desc, int source_image, int* stat, int num_e
...
+}
+
+void c_co_sum(CFI_cdesc_t* a_desc, int result_image, int* stat, char* errmsg, size_t num_elements)
+{
+  CFI_type_t a_type = a_desc->type;
+  size_t c_sizeof_a;
+
+  switch (a_type)
+  {
+    case CFI_type_int32_t:        a_type = GEX_DT_I32; break;
```
bonachea, line 156: `gex_DT_t` is not guaranteed to be any particular type and in general will not match `CFI_type_t`, which means the assignment above only happens to compile.

Please declare a second variable of type `gex_DT_t` and use that to hold GEX_DT constants.

src/caffeine/caffeine.c
```diff
@@ -157,9 +137,61 @@ void c_co_broadcast(CFI_cdesc_t * a_desc, int source_image, int* stat, int num_e
...
+#if __GNUC__ >= 12
+  if (a_type == CFI_type_float_Complex || a_type == CFI_type_double_Complex) 
+#else
+  if (a_type == 2052 || a_type == 4100)
+#endif
+  {
+    c_sizeof_a = a_desc->elem_len/2;
+  } else {
+    c_sizeof_a = a_desc->elem_len;
+  }
```
bonachea, line 185: Suggest folding the above logic into the switch statement by initializing `c_sizeof_a = a_desc->elem_len;` and then setting it to the smaller value for the two complex cases.

src/caffeine/caffeine.c
```diff
@@ -157,9 +137,61 @@ void c_co_broadcast(CFI_cdesc_t * a_desc, int source_image, int* stat, int num_e
...
+    case CFI_type_int64_t:        a_type = GEX_DT_I64; break;
+    case CFI_type_float:          a_type = GEX_DT_FLT; break;
+    case CFI_type_double:         a_type = GEX_DT_DBL; break;
+#if __GNUC__ >= 12
+    case CFI_type_float_Complex:  a_type = GEX_DT_FLT; num_elements *= 2; break;
+    case CFI_type_double_Complex: a_type = GEX_DT_DBL; num_elements *= 2; break;
+#else
+    case 2052:                    a_type = GEX_DT_FLT; num_elements *= 2; break;
+    case 4100:                    a_type = GEX_DT_DBL; num_elements *= 2; break;
+#endif
```
bonachea, line 166: Ick to these constants.

I suggest #defining macros called CAF_CFI_type_{float,double}_Complex at the top of this file and use those here.

src/caffeine/caffeine.c
```diff
@@ -157,9 +137,61 @@ void c_co_broadcast(CFI_cdesc_t * a_desc, int source_image, int* stat, int num_e
...
+    default:
+      if (stat == NULL && errmsg == NULL) {
+        printf("caffeine.c: unrecognized type %d\n", a_type);
+        abort();
```
bonachea, line 170: I suggest replacing the two lines above with `gasnett_fatalerror("caffeine.c: unrecognized type %d\n", a_type)`. Similarly anywhere else you are issuing a fatal error.  This will give you much nicer crash behavior than raw `abort()`.

Docs:

void gasnett_fatalerror(const char *msg, ...);
void gasnett_fatalerror_nopos(const char *msg, ...);

  Issue an fatal error message to the console, as specified by the arguments
  (which follow a printf format convention). Then freeze for debugger and/or
  print a backtrace (depending on current settings) and issue an abort().
  The first form includes information about call site (i.e. function name,
  source file and line number), whereas the second omits this information.

bonachea (changes requested):
Overall looks like a great improvement!

Added a few suggestions based on quick skim.

src/caffeine/caffeine.c
rouson: Note to self: use two different variables: `a_type` and something to hold the `GEX_DT_*` value.  The latter should be of type `gex_DT_t`.

src/caffeine/caffeine.c
rouson: Will do.

src/caffeine/caffeine.c
rouson: fixed by fe23297dc64045e934160a328a03a38afb3eba59

src/caffeine/caffeine.c
rouson: fixed bevy 6caf2284db00d65000b54cbc989386d02df26914

PR #24, opened 2022-03-01
Add normal termination & fix error termination

rouson:
This PR
1. Fixes the error termination behavior in `caf_error_stop` by removing a synchronization and
2. Adds normal termination support via `caf_stop`, including the standard-conforming synchronization.


README.md
```diff
@@ -45,14 +63,24 @@ Run tests
 
 Generate documentation
 ----------------------
+Generate HTML documentation for Caffeine using [ford]:
 ```
 ford doc-generator.md
 ```
 Open `doc/html/index.htmtl` in a web browser.
```
bonachea, line 70: `index.htmtl` is probably a typo

rouson: Good catch!  Thanks, @bonachea.  I'll push a fix momentarily. BTW, I finally bumped into the problem you predicted with the way I was setting `dt_sz`.  Commit 24a6b31252a0073b46c8ded810b54c8d83ad77b1 fixes it.

PR #31, opened 2022-06-14
Adjust compiler detection logic in install script

### LINKED ISSUE #28: Compiler mismatch and full paths
Just got an installer mismatch:
```
$ env CC=gcc CXX=g++ FC=gfortran ./install.sh 
PREFIX=<redacted>/bonachea/.local
PKG_CONFIG_PATH=<redacted>/bonachea/.local/lib/pkgconfig
GASNET_CC=/usr/local/pkg/gcc/11.2.0/bin/gcc and FPM_CC=gcc don't match
```

The workaround:
```
$ env CC=`which gcc` CXX=`which g++` FC=`which gfortran` ./install.sh
```
but should not be necessary...

GASNET_CC is usually absified automatically by configure.
install.sh should absify both commands (e.g. via `type -p $FPM_CC`) before comparing them.
Arguably you should also absify the `$FPM_FC` value you save, to ensure proper operation after a `$PATH` change.

wyphan:
Hopefully closes #28.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -237,8 +237,8 @@ else
 fi
 echo "PKG_CONFIG_PATH=$PKG_CONFIG_PATH"
 
-FPM_FC="$FC"
-FPM_CC="$CC"
+FPM_FC="$($REALPATH $(command -v $FC))"
+FPM_CC="$($REALPATH $(command -v $CC))"
 
 ask_package_permission()
 {
--- a/install.sh
+++ b/install.sh
@@ -91,7 +91,7 @@ if command -v make > /dev/null 2>&1; then
 fi
 
 if command -v fpm > /dev/null 2>&1; then
-  FPM=`which fpm`
+  FPM=`$which fpm`
 fi
 
 ask_permission_to_use_homebrew()
@@ -304,8 +304,9 @@ fi
 # Strip compiler flags
 # Warning: This assumes the full path doesn't contain any spaces!
 GASNET_CC_STRIPPED="$(echo $GASNET_CC | awk '{print $1};')"
+GASNET_CC_REAL="$REALPATH $GASNET_CC_STRIPPED"
 
-if [ "$GASNET_CC_STRIPPED" != "$FPM_CC" ]; then 
+if [ "$GASNET_CC_REAL" != "$FPM_CC" ]; then 
   echo "GASNET_CC=$GASNET_CC_STRIPPED" and  "FPM_CC=$FPM_CC don't match"
   exit 1;
 fi
--- a/install.sh
+++ b/install.sh
@@ -304,7 +304,7 @@ fi
 # Strip compiler flags
 # Warning: This assumes the full path doesn't contain any spaces!
 GASNET_CC_STRIPPED="$(echo $GASNET_CC | awk '{print $1};')"
-GASNET_CC_REAL="$REALPATH $GASNET_CC_STRIPPED"
+GASNET_CC_REAL="$($REALPATH $GASNET_CC_STRIPPED)"
 
 if [ "$GASNET_CC_REAL" != "$FPM_CC" ]; then 
   echo "GASNET_CC=$GASNET_CC_REAL" and  "FPM_CC=$FPM_CC don't match"
--- a/install.sh
+++ b/install.sh
@@ -297,7 +297,8 @@ if [ $GASNET_IS_SPACK ]; then
   echo "unit-testing purposes, and is generally UNSUITABLE FOR PRODUCTION USE."
   echo "The RECOMMENDED way to build GASNet is as an embedded library as configured"
   echo "by the higher-level client runtime package (i.e. Caffeine), including"
-  echo "system-specific configuration."
+  echo "system-specific configuration. Exiting install.sh"
+  exit 1
 fi
 
 # Strip compiler flags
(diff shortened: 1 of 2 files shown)


bonachea (changes requested):

install.sh
```diff
@@ -289,6 +289,9 @@ GASNET_CC="`$PKG_CONFIG $pkg --variable=GASNET_CC`"
 GASNET_CFLAGS="`$PKG_CONFIG $pkg --variable=GASNET_CFLAGS`"
 GASNET_CPPFLAGS="`$PKG_CONFIG $pkg --variable=GASNET_CPPFLAGS`"
 
+# Strip compiler flags
+GASNET_CC="$(echo $GASNET_CC | awk '{print $1};')"
```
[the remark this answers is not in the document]
wyphan: Yeah that makes sense. Lemme adjust it into a new var.

wyphan:
I tested it on Rice's internal machine (ufront) and it works. Btw, would you mind approving the workflow CI runs?

wyphan:
Welp, I'm not familiar enough to debug the macOS CI run, but it seems like a genuine failure, since fpm was installed through Homebrew while GASNet-Ex was not?

bonachea:
> Welp, I'm not familiar enough with Homebrew to debug the macOS CI run, but it seems like a genuine failure, since fpm was installed through Homebrew while GASNet-Ex was not?

I agree this looks like a genuine mismatch in the CI setup, which is exactly the kind of problem the check is designed to detect.

@rouson will need to look at the CI setup (and review this PR)

install.sh
```diff
@@ -289,8 +289,23 @@ GASNET_CC="`$PKG_CONFIG $pkg --variable=GASNET_CC`"
...
-  echo "GASNET_CC=$GASNET_CC" and  "FPM_CC=$FPM_CC don't match"
+# Check whether GASNet was installed using Spack and print warning message
+# Note: relies on the fact that most Spack installations have "opt/spack" in the directory path
+GASNET_IS_SPACK=$(echo $GASNET_LIBS | grep -c -v spack)
+if [ $GASNET_IS_SPACK ]; then
+  echo "***NOTICE***: The GASNet library built by Spack is ONLY intended for"
+  echo "unit-testing purposes, and is generally UNSUITABLE FOR PRODUCTION USE."
+  echo "The RECOMMENDED way to build GASNet is as an embedded library as configured"
+  echo "by the higher-level client runtime package (i.e. Caffeine), including"
+  echo "system-specific configuration."
```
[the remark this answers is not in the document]
rouson:
```suggestion
  exit 1
```

rouson (changes requested):

install.sh
```diff
@@ -91,7 +91,7 @@ if command -v make > /dev/null 2>&1; then
 fi
 
 if command -v fpm > /dev/null 2>&1; then
-  FPM=`which fpm`
+  FPM=`$which fpm`
```
bonachea, line 94: What is the significance of this change?

wyphan: Whoops, that's a typo. I think I put $($REALPATH fpm)` there and realized that's not needed, but forgot to remove it completely.

install.sh
```diff
@@ -289,8 +289,25 @@ GASNET_CC="`$PKG_CONFIG $pkg --variable=GASNET_CC`"
 GASNET_CFLAGS="`$PKG_CONFIG $pkg --variable=GASNET_CFLAGS`"
 GASNET_CPPFLAGS="`$PKG_CONFIG $pkg --variable=GASNET_CPPFLAGS`"
 
-if [ "$GASNET_CC" != "$FPM_CC" ]; then 
-  echo "GASNET_CC=$GASNET_CC" and  "FPM_CC=$FPM_CC don't match"
+# Check whether GASNet was installed using Spack. If yes, bail out.
+# Note: relies on the fact that most Spack installations have "opt/spack" in the directory path
+GASNET_IS_SPACK=$(echo $GASNET_LIBS | awk '{print $1};' | grep -c -v spack)
```
bonachea, line 294: The comment should probably also note that the command above assumes the ordering of `$GASNET_LIBS` places the GASNet library directory first, which is currently true but not guaranteed (although also not likely to change).

However I don't understand the `-v` option to `grep` - isn't that backwards?

wyphan: The `-v` option inverts the count such that 0 is true because I thought that's what POSIX `test` / `[` expects (so I can simply put `if [ $cond ]`. But I guess I can remove that since I explicitly test for equality using `-eq`.

bonachea (approved):
Looks right to me. 

We're just waiting for @rouson to take a look at why CI is failing.

install.sh
```diff
@@ -289,8 +289,27 @@ GASNET_CC="`$PKG_CONFIG $pkg --variable=GASNET_CC`"
...
 GASNET_CPPFLAGS="`$PKG_CONFIG $pkg --variable=GASNET_CPPFLAGS`"
 
-if [ "$GASNET_CC" != "$FPM_CC" ]; then 
-  echo "GASNET_CC=$GASNET_CC" and  "FPM_CC=$FPM_CC don't match"
+# Check whether GASNet was installed using Spack. If yes, bail out.
+# Note: relies on the fact that most Spack installations have "opt/spack"
+#       in the directory path, and assumes that the first directory returned
+#       by pkg-config contains the GASNet lib directory
+GASNET_LIBDIR="$(echo $GASNET_LIBS | awk '{print $1};')"
+if [ "$(echo $GASNET_LIBDIR | grep -q spack)" ]; then
```
rouson, line 297: @wyphan let's use `sh` [built-in glob-matching] rather than `grep` to avoid forking

[built-in glob-matching]: https://stackoverflow.com/a/23472637

rouson (changes requested):

rouson (approved):
LGTM

PR #34, opened 2022-07-27
chore: vendor the assert library

everythingfunctional:
Vendor the assert library to avoid potential circular dependency.


rouson (approved):
LGTM

PR #37, opened 2022-08-03
fix(co_reduce test): standardize logicals

rouson:


rouson:
reviewed by @everythingfunctional

PR #38, opened 2022-08-09
Update  dependencies: GCC, GASNet-EX, Veggies, and Cart

rouson:


install.sh
```diff
@@ -18,7 +18,7 @@ print_usage_info()
     echo "yes | ./install.sh"
 }
 
-GCC_VERSION=11
+GCC_VERSION=12
 GASNET_VERSION="2021.9.0"
```
bonachea, line 22: How about also updating GASNet to the current 2022.3.0 release?

bonachea (changes requested):

install.sh
rouson: @bonachea done.  In case it's useful, you can make a suggested edit in a comment by clicking the boxed "+/" button and then I can click to accept and apply the change.

everythingfunctional:
My guess at the moment is that the homebrew installed fpm is linked to gcc-11, but you only installed gcc-12. Thus, the error from the fpm command. I will look into it further and let you know what I'd recommend.

everythingfunctional:
I'd recommend just having the github action install fpm in the CI. It appears to work the most reliably. The downside is now you're not checking in the CI that your install script can install fpm. I'd say it's not crucial to do so, as there's not really anything complicated about how you're doing it, and the chances are pretty small that many people will rely on this script for that. My 2 cents.

rouson:
Thanks @everythingfunctional.  All tests pass so I'll merge this.

PR #39, opened 2022-09-27
Fix installer Linux

rouson:
On Linux, a recent Homebrew [change] requires executing `brew force-link glibc` after `brew link gcc`.  This PR adds this command where necessary in `install.sh`.

[change]: https://github.com/actions/runner-images/issues/6280#issuecomment-1257147693


everythingfunctional:
I couldn't find a way in the CI to get homebrew:
* installed
* usable for installing dependencies, and
* have the packages installed by homebrew work

So (for now at least), I just used the Linux package manger to install the dependencies so we can at least build and run the tests for Caffeine.

PR #40, opened 2022-12-19
Fix missing bracket

awehrfritz:
This PR removes a leftover bracket in:
https://github.com/BerkeleyLab/caffeine/blob/38f2fd2c8d8017af37f74daeb85d1a5da04a23b9/src/caffeine/caffeine.c#L244

I also corrected a few minor typos as I was going through the README and documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ Paper: <https://doi.org/10.25344/S4459B>
 
 Documentation
 -------------
-One of our continous integration (CI) scripts generates up-to-date Caffeine documentation using [ford].  The CI script also deploys the generated documentation to the our GitHub Pages [site].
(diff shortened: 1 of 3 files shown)


rouson (approved):
LGTM

rouson:
@awehrfritz thanks for this pull request.  I'm very sorry about missing it until now.   At the suggestion of the LLVM Flang developer community, we have been focused on writing a parallel runtime interface design document.  The goal is for any compiler that targets the interface to be able to use any runtime library that supports the interface.  With the design document close to being posted for comments, we are resuming work on Caffeine to expand its support for the new interface.  Our first goal is to begin supporting coarray allocations.

PR #41, opened 2023-01-10
Add beginnings of `rma` module

ktras:
Add the beginnings of a module to add gets and puts. This was drafted collectively in a meeting and have now been translated into compilable code. The name and placement of the module are just suggestions and very much open to feedback.

### DIFF
--- a/src/caffeine/rma_m.f90
+++ b/src/caffeine/rma_m.f90
@@ -1,0 +1,39 @@
+! Copyright (c), The Regents of the University of California
+! Terms of use are as specified in LICENSE.txt
+module rma_m
+  use team_type_m, only: team_type
+  implicit none
+
+  private
+  public :: caf_put, caf_get
+
+  type caf_co_handle
+  end type
+
+  interface
+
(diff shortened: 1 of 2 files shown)


bonachea:
Hi @ktras Thanks for creating this PR!

I'll let @rouson comment on source file naming/placement, I don't have a strong opinion on that.

I'm fine with this file living in the Caffeine repository, but in the near-term I think it's much more important for this interface work to "grow upwards" to flesh out the interface design and semantics (the part that needs near-term community buy-in), rather than "growing downwards" into implementation details (which might not be fully usable for months, and doesn't require any consensus beyond our group). In particular, the next thing I'd like to see added is more detailed descriptions of the subroutine arguments and client-facing behavioral semantics of the subroutine(s), ideally along with a usage example. Do we want to build that kind of documentation in source code comments and later extract it with something like FORD? Or should we just write it directly into a markdown design document?

ktras:
Out of date with current work elsewhere that will be integrated into this repo in the future. Closing the PR.

PR #43, opened 2023-12-07
chore(install.sh): update to GCC 13

rouson:


install.sh
```diff
@@ -18,7 +18,7 @@ print_usage_info()
     echo "yes | ./install.sh"
 }
 
-GCC_VERSION=12
+GCC_VERSION=13
 GASNET_VERSION="2022.3.0"
```
bonachea, line 22: Weak request to also update the GASNet version

ktras: @bonachea Is the above suggestion the most recent version? Or at least a recent version you would recommend?

bonachea: Yes that remains the current release version until we finish our new release later this month.

PR #46, opened 2023-12-13
Update interfaces

ktras:
Update interfaces based on latest design decisions.

### DIFF
--- a/src/caffeine/assert/assert_s.f90
+++ b/src/caffeine/assert/assert_s.f90
@@ -5,6 +5,7 @@
 !     contract # NRC-HQ-60-17-C-0007
 !
 submodule(caffeine_assert_m) caffeine_assert_s
+  use iso_c_binding, only: c_bool
   implicit none
 
 contains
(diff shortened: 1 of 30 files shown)


src/caffeine/assert/assert_s.f90
```diff
@@ -52,7 +53,7 @@
 
         end if represent_diagnostics_as_string
 
-        call prif_error_stop(header // ' with diagnostic data "' // trailer // '"')
+        call prif_error_stop(logical(.false., c_bool), stop_code_char=header // ' with diagnostic data "' // trailer // '"')
```
everythingfunctional, line 56: Note that you can use `.false._c_bool` instead of `logical(.false., c_bool)`.

everythingfunctional (commented):
Aside from the small comment, looks good to me

src/caffeine/assert/assert_s.f90
ktras: I tried that and I got a compiler error with `gfortran`. I will try it again and copy the error here if it happens again.

everythingfunctional: Interesting. The standard says

> R725 *logical-literal-constant* **is** .TRUE. [ _ *kind-param* ]
>                                               **or** .FALSE. [ _ *kind-param* ]

ktras: Actually, no compiler error. I think the first time I tried that syntax here, I must not have had the `c_bool` `use-stmt` necessary and the compiler error ended up being a confusing one that made me think the syntax wasn't supported. I have updated all occurrence with your suggested syntax, since I also like it better.

everythingfunctional (approved):

PR #47, opened 2023-12-15
Add `prif_allocate`, `caf_allocate`, and `dlmalloc`

ktras:
Add `prif_allocate`, `caf_allocate`, and add files for `dlmalloc`. `caf_allocate` currently has a trivial implementation that uses the system `malloc` to grab memory, but will be replaced by `dlmalloc`. This PR does not yet use `dlmalloc` and such work will be added in a future PR.


rouson (approved):
LGTM

PR #48, opened 2023-12-18
Add all remaining `prif` interfaces

ktras:
Add all remaining `prif` interfaces based on design doc revision 0.2

### DIFF
--- a/src/prif.f90
+++ b/src/prif.f90
@@ -5,7 +5,7 @@ module prif
   use program_termination_m, only : prif_stop, prif_error_stop, prif_fail_image
   use allocation_m, only: prif_coarray_handle, prif_allocate
   use coarray_queries_m, only: prif_lcobound, prif_ucobound, prif_coshape, prif_image_index
-  use image_queries_m, only : prif_this_image, prif_num_images
+  use image_queries_m, only : prif_this_image, prif_num_images, prif_failed_images, prif_stopped_images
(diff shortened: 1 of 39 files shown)


src/prif.f90
bonachea, line 1: Missing new `prif_critical_type`, which should probably be just a renamed `prif_lock_type`

bonachea (changes requested):
Overall LGTM. 

I did NOT attempt to exhaustively compare all the subroutine signatures for spec conformance, although that's something that eventually needs to be audited, and ideally supported with test code that ensures calls compile as intended.

everythingfunctional (approved):
A did a brief once-over and nothing jumps out. LGTM

src/prif.f90
ktras: Done.

bonachea (approved):
Removing `pure` from functions internal to caffeine is probably overkill, but I have no specific objection.

The important part is to match the specified `prif_` interfaces at the top.

ktras:
> Removing `pure` from functions internal to caffeine is probably overkill, but I have no specific objection.
> 
> The important part is to match the specified `prif_` interfaces at the top.

@bonachea, I agree. I only removed `pure` from functions that I had to based on the cascading affects of `pure` being removed from some of the `prif_` interfaces that called other `pure` functions and no longer can. Thanks for the review!

PR #52, opened 2023-12-21
Use dlmalloc for coarray allocation

everythingfunctional:
Provides a basic implementation for allocating and deallocating coarrays, with some preliminary consideration for teams.

### DIFF
--- a/src/caffeine/allocation_m.f90
+++ b/src/caffeine/allocation_m.f90
@@ -6,6 +6,7 @@ module allocation_m
     implicit none
     private
     public :: &
+        handle_data, &
         prif_coarray_handle, &
         prif_allocate, &
         prif_allocate_non_symmetric, &
--- a/src/caffeine/allocation_m.f90
+++ b/src/caffeine/allocation_m.f90
@@ -2,13 +2,33 @@
 ! Terms of use are as specified in LICENSE.txt
 module allocation_m
     use iso_c_binding, only: c_ptr, c_int, c_intmax_t, c_size_t, c_funptr
+
     implicit none
     private
-    public :: prif_allocate, prif_allocate_non_symmetric, prif_deallocate, prif_deallocate_non_symmetric
+    public :: &
+        prif_coarray_handle, &
+        prif_allocate, &
+        prif_allocate_non_symmetric, &
+        prif_deallocate, &
+        prif_deallocate_non_symmetric
 
+    ! TODO: Should these actually be interleaved if we use explicit size?
+    type, bind(C) :: cobound_pair
+      integer(c_intmax_t) :: lcobound, ucobound
+    end type
+    
+    type, bind(C) :: handle_data
+      type(c_ptr) :: coarray_data
+      integer(c_int) :: corank
+      integer(c_size_t) :: coarray_size
+      type(c_funptr) :: final_func
+      type(c_ptr) :: previous_handle, next_handle
+      type(cobound_pair) :: cobounds(15)
+    end type
 
-    type, public :: prif_coarray_handle
-       type(c_ptr) :: ptr
+    type :: prif_coarray_handle
+      private
+      type(handle_data), pointer :: info
     end type
 
     interface
--- a/src/caffeine/allocation_m.f90
+++ b/src/caffeine/allocation_m.f90
@@ -60,7 +60,8 @@ module allocation_m
 
        module subroutine prif_deallocate(coarray_handles, stat, errmsg, errmsg_alloc)
          implicit none
-         type(prif_coarray_handle), intent(in) :: coarray_handles(:)
+         ! TODO: I think we need to add target here, to be able to pass to bind(C) final_func
+         type(prif_coarray_handle), intent(inout) :: coarray_handles(:)
          integer(c_int), intent(out), optional :: stat
          character(len=*), intent(inout), optional :: errmsg
          character(len=:), intent(inout), allocatable, optional :: errmsg_alloc
--- a/src/caffeine/allocation_s.f90
+++ b/src/caffeine/allocation_s.f90
@@ -3,6 +3,7 @@
 submodule(allocation_m) allocation_s
   use iso_c_binding, only: c_sizeof, c_f_pointer, c_loc, c_null_ptr
   use caffeine_h_m, only: caf_allocate
+  use synchronization_m, only: prif_sync_all
   use teams_m, only: prif_team_type, current_team
 
   implicit none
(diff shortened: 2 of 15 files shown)


src/caffeine/allocation_m.f90
```diff
@@ -4,11 +4,19 @@ module allocation_m
...
+        prif_deallocate_non_symmetric
 
 
-    type, public :: prif_coarray_handle
-       type(c_ptr) :: ptr
+    type :: prif_coarray_handle
+       integer(c_intmax_t), allocatable :: lcobounds(:), ucobounds(:)
+       integer(c_size_t) :: element_length
+       type(c_funptr) :: final_func
+       type(c_ptr) :: base_pointer
```
bonachea, line 19: This is almost certainly not the approach we want. 

These all need to be stored in the heap as a metadata header before the elements (as in Kate's prototyping), not in the handle itself.

bonachea (changes requested):

src/caffeine/allocation_m.f90
```diff
@@ -2,13 +2,33 @@
...
+      integer(c_intmax_t) :: lcobound, ucobound
+    end type
+    
+    type, bind(C) :: handle_data
+      type(c_ptr) :: coarray_data
+      integer(c_int) :: corank
+      integer(c_size_t) :: coarray_size
+      type(c_funptr) :: final_func
+      type(c_ptr) :: previous_handle, next_handle
+      type(cobound_pair) :: cobounds(15)
```
bonachea, line 26: TODO: Missing `type(c_ptr) :: context_data`

Probably also missing a pointer to the establishment team, once that type is defined

Otherwise looks like a good start!

everythingfunctional: Actually, I don't think a coarray does need a pointer to the establishing team. All coarray accesses are either with respect to the current team (which is easily grabbed), or with respect to an explicitly provided team.

bonachea: Are you sure? What about a coarray established in a parent team and not reassociated in the CHANGE TEAM statement? Aren't coindexed expressions on such an array (with no explicit team) interpreted wrt the establishing team?

everythingfunctional: > If a TEAM= specifier appears in an image-selector ...  If a TEAM_NUMBER= specifier appears in an image-selector ... Otherwise, the team of the image selector is the current team.

> An image selector shall specify an image index value that is not greater than the number of images in the team of the image selector, and identifies the image with that index in that team.

src/caffeine/caffeine.c
```diff
@@ -22,126 +21,40 @@ static gex_Rank_t rank, size;
...
-
-#define PTR_DIFFERENCE_BYTES(first_element_addr, fortran_object_base_addr) ((char*)(first_element_addr) - (char*)(fortran_object_base_addr))
-
-static void setupCoarrayHandle(void* handle_mem, int corank, intmax_t* lcobounds, intmax_t* ucobounds, final_func_ptr final_func, size_t sz, int64_t allocator, void* object_base_addr_ptr);
-
-void caf_caffeinate(int argc, char *argv[])
+void caf_caffeinate(mspace* symmetric_heap, gex_TM_t** initial_team)
 {
-  GASNET_SAFE(gex_Client_Init(&myclient, &myep, &myteam, "caffeine", &argc, &argv, 0));
+  *initial_team = malloc(sizeof(gex_TM_t));
```
bonachea, line 26: `gex_TM_t` is a scalar type (a pointer to opaque data), and acts as a trivially-copyable handle.
I would prefer we pass it around by value instead of adding this extra level of indirection on `gex_TM_t`.

If we need an interoperable opaque type, we can safely reinterpret cast a `gex_TM_t` to a `c_intptr_t` or `c_ptr` in the same way we are currently doing for `mspace`.

bonachea (changes requested):

src/caffeine/allocation_m.f90
bonachea: >  Otherwise, the team of the image selector is the current team.

I don't understand how the coindexes are interpreted wrt to a current team that is a subset of the establishing team when corank > 1. However I assume this is explained somewhere..

src/caffeine/caffeine.c
everythingfunctional: It does seem to work, but just to be sure we understand whether it's supposed to... The reason it works for the `mspace` is because `mspace` is defined as `typedef void* mspace`. Is `gex_TM_t` defined somewhat similarly? I.e. it's guaranteed to have the same size as a `void*`?

bonachea: It works for the same reason: `gex_TM_t` is implemented as a C pointer type (a pointer to an incomplete type, similar to `void *`, and guaranteed by C to have the same size).

`gex_TM_t` is not technically guaranteed by specification to be a pointer type, but that's the current implementation and it's very unlikely to ever change. Our UPC++ implementation is also currently relying on this property, so I have no hesitation recommending the same here.

everythingfunctional:
@bonachea , if you want to give this one more look and then merge, I'll start working on teams support in a separate PR while @ktras starts looking at put/get implementations.

src/caffeine/allocation_s.f90
```diff
@@ -1,25 +1,154 @@
...
+        if (present(errmsg)) then
+          errmsg = unallocated_message
+        else if (present(errmsg_alloc)) then
+          errmsg_alloc = unallocated_message
+        end if
+        return
+      else
+        call prif_error_stop(.false._c_bool, stop_code_char=unallocated_message)
+      end if
+    end if
```
bonachea, line 90:
```suggestion
    end if
    ! TODO: invoke finalizers from coarray_handles(:)%info%final_func
```
I'd like to hear more about why Fortran is rejecting the commented-out code, but that discussion can take place post-merge. 

We can always invoke the finalizer from the C-side if necessary to circumvent any Fortran-side stupidity.

bonachea (approved):
Requested one additional comment, otherwise LGTM!

src/caffeine/allocation_s.f90
everythingfunctional: It's a bug in gfortran. I've submitted a bug-report. There is a subtle change we'll need to make to the interface of `prif_deallocate` though.

ktras:
@everythingfunctional @bonachea Can this PR be merged now?

bonachea:
Yes, I've already Approved

PR #53, opened 2023-12-30
Default to the GASNet-EX stable tarball

bonachea:
Until we decide to strongly couple Caffeine releases with GASNet ones, it's probably best to default to the GASNet stable tarball, which always contains the latest stable bits.


ktras (approved):
LGTM

PR #54, opened 2024-01-16
have only team leader perform symmetric allocations

everythingfunctional:


src/caffeine/caffeine_h_m.f90
```diff
@@ -154,7 +156,20 @@ pure function caf_elem_len(a) result(a_elem_len) bind(C)
...
+       type(c_ptr), intent(in), value :: ptr
+       integer(c_intptr_t) :: as_int
+     end function
+
+     pure function as_c_ptr(i) bind(C)
+       !! void* as_c_ptr(intptr_t i);
+       import c_ptr, c_intptr_t
+       integer(c_intptr_t), intent(in), value :: i
+       type(c_ptr) :: as_c_ptr
+     end function
```
bonachea, line 172: Is a library function call into C code really the only way for Fortran code to convert between `c_ptr` and `c_intptr_t`? 

If so that's really unfortunate because it takes what should essentially be an instruction-free type conversion and adds the full overhead of a cross-object function call that is unlikely to be removed by anything short of full LTO.

The call sites added in this PR occur on a path that includes a collective operation, so the call overhead is likely lost in the noise. However I'm mildly concerned that if these conversions creep into other non-collective fast-paths, the overhead might become noticeable. Not requesting a change at this time, just something to watch out for.

bonachea (approved):
Added one comment, otherwise LGTM

src/caffeine/caffeine_h_m.f90
everythingfunctional: Not in a standards conforming way, and not in a way that the compiler will tell you when it's not actually valid. The Fortran standard goes a long way to avoid saying that "pointers are integer values representing memory addresses", and static casting is not a thing in Fortran. That said you could do one of the following

```Fortran
type(c_ptr), target :: ptr, new_ptr
integer(c_intptr_t), pointer :: ptr_as_int, new_ptr_as_int
call c_f_pointer(c_loc(ptr), ptr_as_int)
call c_f_pointer(c_loc(new_ptr), new_ptr_as_int)
new_ptr_as_int = ptr_as_int + 1
```
or
```Fortran
type(c_ptr) :: ptr, new_ptr
integer(c_intptr_t) :: ptr_as_int, new_ptr_as_int
ptr_as_int = transfer(ptr, ptr_as_int)
new_ptr_as_int = ptr_as_int + 1
new_ptr = transfer(new_ptr_as_int, new_ptr)
```
The library call is effectively equivalent to the second one, but with still some amount of type-checking.

PR #56, opened 2024-02-07
Add to `prif_image_index` ability to handle cases of invalid cosubscripts

ktras:
Add to `prif_image_index` ability to handle cases of invalid cosubscripts by returning zero. Add case in image_index test

### DIFF
--- a/src/caffeine/coarray_queries_s.f90
+++ b/src/caffeine/coarray_queries_s.f90
@@ -1,6 +1,7 @@
 ! Copyright (c), The Regents of the University of California
 ! Terms of use are as specified in LICENSE.txt
 submodule(coarray_queries_m) coarray_queries_s
+  use image_queries_m, only: prif_num_images
 
   implicit none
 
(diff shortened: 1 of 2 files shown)


ktras:
This functionality breaks the more complicated test for `image_index`. I am not sure how to update that test so that the complicated `image_index` case can still be tested successfully when not actually running on enough nodes.

src/caffeine/coarray_queries_s.f90
```diff
@@ -22,15 +23,32 @@
...
+          exit check_subscripts
+       end if
+    end do check_subscripts
+
+    if (.not. invalid_cosubscripts) then
+      image_index = 1 + sub(1) - coarray_handle%info%lcobounds(1)
+      prior_size = 1
+      do dim = 2, size(sub)
+        prior_size = prior_size * (coarray_handle%info%ucobounds(dim-1) - coarray_handle%info%lcobounds(dim-1) + 1)
+        image_index = image_index + (sub(dim) - coarray_handle%info%lcobounds(dim)) * prior_size
```
bonachea, line 44: The above calculation looks wrong to me.

Have you tested anything with corank > 1?

PR #57, opened 2024-02-07
Initial implementations of the contiguous RMA operations

everythingfunctional:


src/caffeine/coarray_queries_s.f90
```diff
@@ -22,6 +23,32 @@
...
+       end if
+    end do check_subscripts
+
+    if (.not. invalid_cosubscripts) then
+      image_index = 1 + sub(1) - coarray_handle%info%lcobounds(1)
+      prior_size = 1
+      do dim = 2, size(sub)
+        prior_size = prior_size * (coarray_handle%info%ucobounds(dim-1) - coarray_handle%info%lcobounds(dim-1) + 1)
+        image_index = image_index + (sub(dim) - coarray_handle%info%lcobounds(dim)) * prior_size
+       end do
```
bonachea, line 45:
```suggestion
       end do
       ```suggestion
      ! Future work: values of prior_size are invariant across calls w/ the same coarray_handle
      !  We could store them in the coarray metadata at allocation rather than redundantly 
      ! computing them here, which would accelerate calls with corank > 1 by removing 
      ! corank multiply/add operations and the loop-carried dependence
      prior_size = 1
```
```

Comment (everythingfunctional) [reply src/caffeine/coarray_queries_s.f90:49]: Agree. I've added the comment.

**Comment by everythingfunctional** [comment]
I think this is ready now. I'll note that I had to add the `target` attribute to the `value` arguments to be able to call `c_loc` on them. (I also submitted a PR to the design doc: https://github.com/ktras/prif/pull/55). It doesn't add any constraints to the caller, just a signal to the compiler not to do certain optimizations with the argument.

File: src/caffeine/caffeine_h_m.f90 [hunk 2]
```diff
@@ -149,13 +171,13 @@ logical(c_bool) pure function caf_same_cfi_type(a, b) bind(C)
...
      end function
-  
+
      logical(c_bool) pure function caf_numeric_type(a) bind(C)
        !! bool caf_numeric_type(CFI_cdesc_t* a_desc);
        import c_bool
        type(*), intent(in) :: a(..)
      end function
-  
+
```
bonachea, line 180: These types of whitespace changes in lines that are not otherwise modified by a PR have the potential to slow down review and create spurious merge conflicts. Streamlined teamwork is far more important than a few stray/meaningless whitespace characters.

I assume these whitespace changes were performed automatically by your editor or some script. I highly encourage both @everythingfunctional and @ktras to DISABLE any such features that automatically create spurious whitespace changes over any files you happen to edit, to avoid these problems. I would even go so far as adding this to the coding standards; as an example, here are some of the relevant portions of the [GASNet coding standard](https://bitbucket.org/berkeleylab/gasnet/src/stable/README-devel):

```
* Do NOT use tab characters in GASNet code files. Use spaces to ensure correct
  formatting in any editor.
    Emacs users should: (setq-default indent-tabs-mode nil)
    Vim users should: :set expandtab

* Do NOT commit gratuitous changes to whitespace, even to remove tab characters.
  Such changes have no semantic content and can make use of revision history more
  difficult.  You can preview your changes with the following two commands:
     git diff -u | wc -l
     git diff -u -b | wc -l
  and if the outputs differ, then your changes include lines that have changed
  in only the whitespace characters.

* Large code motion (more than a dozen lines mostly unchanged) should be
  isolated in a commit by itself, whenever practical.

...
```

src/caffeine/coarray_access_s.f90
```diff
@@ -1,24 +1,66 @@
...
   module procedure prif_get
+    integer(c_int) :: image
+    integer(c_intptr_t) :: remote_base, remote_ptr
+
+    call prif_image_index(coarray_handle, coindices, image_index=image)
+    call prif_base_pointer(coarray_handle, image, remote_base)
+    remote_ptr = &
+        remote_base &
+        + (as_int(first_element_addr) &
+        - as_int(coarray_handle%info%coarray_data))
```
bonachea, line 50: As previously expressed, I'm concerned about the superfluous performance overhead incurred by the `as_int()` helper invoked in this critical-path. As currently written, `prif_get()` requires four cross-language calls (prif_base_pointer->caf_convert_base_addr, 2x as_int and prif_get_raw->caf_get). All such calls should be expected to incur a full function call overhead at runtime and inhibit optimization of surrounding code. Ideally we'd replace `as_int` with a pure Fortran helper that the Fortran compiler can analyze and optimize the calls down to simple arithmetic.

In the absence of that ideal solution, we can still halve the number of  function call overheads by eliminating the calls to `as_int`: adding a C-side wrapper around `caf_get` that accepts the three pointer-sized arguments and performs the arithmetic to compute `remote_ptr` before calling the raw `caf_get`.

bonachea (changes requested):
Added one cosmetic observation and one technical issue for discussion.

src/caffeine/caffeine_h_m.f90
everythingfunctional: Apologies for the accidental whitespace changes. I've undone them.

src/caffeine/coarray_access_s.f90
bonachea: This is now entered as issue #67 which proposes to resolve these using the "ideal solution" mentioned above.

bonachea (approved):

ktras (approved):
LGTM

PR #58, opened 2024-02-12
Add `prif_base_pointer`

ktras:


src/caffeine/caffeine.c
```diff
@@ -26,20 +28,19 @@ void caf_caffeinate(mspace* symmetric_heap, intptr_t* symmetric_heap_start, mspa
   GASNET_SAFE(gex_Client_Init(&myclient, &myep, initial_team, "caffeine", NULL, NULL, 0));
```
bonachea, line 28: The `gex_TM_t` representing the initial team is an invariant and trivially copyable handle that doesn't change over the lifetime of the job.

Please add a `gex_TM_t` global C variable and store a copy of `initial_team` here, so that we can avoid needlessly passing this handle back and forth across the C/Fortran interface.

src/caffeine/caffeine.c
```diff
@@ -90,6 +90,15 @@ void caf_deallocate(mspace heap, void* mem)
   mspace_free(heap, mem);
 }
 
+// take address in a segment and convert to an address on given image
+void* caf_convert_base_addr(void* addr, int image, gex_TM_t initial_team)
```
bonachea, line 94: With the change suggested above, the `initial_team` argument here can and should be removed.

src/caffeine/allocation_m.f90
```diff
@@ -29,7 +29,6 @@ module allocation_m
     end type
 
     type :: prif_coarray_handle
-      private
```
bonachea, line 32: What are the implications of this change? 

I think we should continue to leverage the Fortran type system to strongly discourage PRIF clients from touching the unspecified metadata inside the `prif_coarray_handle`, because we want unfettered flexibility to change things below the PRIF interface. IMO this enforced encapsulation in our library implementation is far more important than enabling open-box testing, which seems to be the point in this commit.

IMHO we should revert this change, and either remove the test or deploy some test-specific workaround for inspecting library-internal state (notably using types and functions that are NOT `prif_`).

src/caffeine/caffeine.h
```diff
@@ -33,6 +33,10 @@ int caf_num_images(gex_TM_t team);
 void* caf_allocate(mspace heap, size_t bytes);
 void caf_deallocate(mspace heap, void* mem);
 
+// PRIF Queries
+
+void* caf_convert_base_addr(void* addr, int image, gex_TM_t initial_team);
```
bonachea, line 38: The Fortran declaration for this `bind(C)` function has a return value of `integer(c_intptr_t)`, but the C definition declares the return type as `void *`.

**values** of these types have convertible representations, but `intptr_t` and `void *` are not guaranteed to have the same in-memory representation (for example, `intptr_t` is permitted to be "wider" than `void *`). Consequently, the Fortran standard correctly does not guarantee these types as interoperable with each other: IIUC this is a mismatched declaration and therefore UB. I believe the C version needs to be declared to return `intptr_t` to ensure guaranteed compatibility.

IIUC neither compiler will report this class of type discrepancies between `bind(C)` Fortran function declarations and the corresponding C function definition. As such, we need to be especially vigilant with manually checking such correspondences. I suggest adding mention of this fact to the coding conventions.

src/caffeine/prif_queries_s.f90
```diff
@@ -13,6 +18,16 @@
   end procedure
 
   module procedure prif_base_pointer
+    integer(c_int) :: num_img
+
+    call prif_num_images(image_count=num_img)
+    call assert(image_num .gt. 0 .or. image_num .lt. num_img, "prif_base_pointer: image_num not within valid range")
```
bonachea, line 24: I believe there are two off-by-one mistakes in the assertion expression above.

Please add tests to detect these defects before fixing them.

src/caffeine/prif_queries_s.f90
```diff
@@ -13,6 +18,16 @@
...
+    if (image_num .eq. 0) then
+      ptr = 0
+    else
+      ! TODO: after more team work, replace last arg with initial_team%gex_team
```
bonachea, line 28: No, this argument should be removed entirely.
```suggestion
```

bonachea (changes requested):
Good start, but made a few important requests for change

src/caffeine/allocation_m.f90
ktras: @bonachea, Dan, I understand your preference. However, it is not just for the tests. I needed to remove that attribute to be able to pass the coarray address to C from `prif_base_pointer` as seen [here](https://github.com/BerkeleyLab/caffeine/pull/58/files#diff-228007d71dbe1e7e0f3f18a32feda012d02b2b9a1c21963b3b036b4e39ec3887R29). Also, Brad's @everythingfunctional [PR](https://github.com/BerkeleyLab/caffeine/pull/57/files#diff-f003feb27e90d53b50e3d227800f63d9e3e19d1014a51c8a19de524841dcc11fL32) also made this change when beginning to work on adding team functionalities, so it was a change I thought was necessary in the long run. Perhaps we should explore this topic more in our meeting this week to see if it is feasible for us to keep it `private`.

bonachea: I agree we should discuss further. 

I suspect the problem is the module structure of caffeine: it probably makes more sense for things like `module allocation_m` to instead be a `submodule` of `module prif`, and that "library-wide internals" like `prif_coarray_handle` and `handle_data` should instead be defined in `module prif` with appropriate `private` qualifiers.

test/caf_prif_queries_test.f90
```diff
@@ -0,0 +1,75 @@
...
+    ucobounds(1) = num_imgs
+    call prif_this_image(image_index=this_img)
+
+    call prif_allocate( &
+      [1_c_intmax_t], ucobounds, [integer(kind=c_intmax_t) ::], [integer(kind=c_intmax_t) ::], 1_c_size_t, c_null_funptr, &
+      coarray_handle, allocated_mem)
+
+    call prif_base_pointer(coarray_handle, this_img, ptr)
+
+    result_ = assert_that(as_int(coarray_handle%info%coarray_data) .eq. ptr)
```
everythingfunctional, line 70: I think you want this to be

```suggestion
    result_ = assert_that(as_int(allocated_mem) .eq. ptr)
```

test/caf_prif_queries_test.f90
```diff
@@ -0,0 +1,75 @@
...
+  function test_prif_queries() result(tests)
+    type(test_item_t) :: tests
+
+    tests = &
+      describe( &
+        "PRIF queries can", &
+        [ it("get a non null base pointer", &
+              check_base_pointer_not_null) &
+        , it("get same address passed into to prif_base_pointer when querying about the same image", &
+              check_same_image_same_addr) &
```
everythingfunctional, line 23: I'd add a test that you do get null for image 0.

src/caffeine/prif_queries_s.f90
```diff
@@ -13,6 +18,16 @@
   end procedure
 
   module procedure prif_base_pointer
+    integer(c_int) :: num_img
+
+    call prif_num_images(image_count=num_img)
+    call assert(image_num .gt. 0 .or. image_num .lt. num_img, "prif_base_pointer: image_num not within valid range")
+    if (image_num .eq. 0) then
+      ptr = 0
```
everythingfunctional, line 26: is it true that (in C) `(intptr_t)NULL == 0`? I'm pretty sure it would definitely be right as `ptr = as_int(c_null_ptr)`

src/caffeine/prif_queries_s.f90
everythingfunctional: To be more specific, I think the "gets null for image 0" covers the first one, and changing the "gets valid pointer" to use `num_img` covers the second.

src/caffeine/allocation_m.f90
everythingfunctional: Yeah, I made it not private because the implementations in other modules needed access to the data, not because the tests did. I think all the coarray accesses/queries (i.e. allocation, RMA, image_index, etc) should go into the module where `prif_coarray_handle` is defined, but we can discuss in more detail tomorrow.

src/caffeine/caffeine.c
```diff
@@ -90,6 +90,15 @@ void caf_deallocate(mspace heap, void* mem)
...
+{
+   size_t offset = (char*)addr - (char*)gex_Segment_QueryAddr(mysegment);
```
everythingfunctional, line 96: Shouldn't this be `ptrdiff_t`? And I always wonder, why `(char*)`?

src/caffeine/prif_queries_s.f90
bonachea: Yes, the C standard guarantees null pointers have the value zero.
6.3.2.3:
> An integer constant expression with the value 0, or such an expression cast to type void *, is called a null pointer constant.66) If a null pointer constant is converted to a pointer type, the resulting pointer, called a null pointer, is guaranteed to compare unequal to a pointer to any object or function.

src/caffeine/caffeine.c
bonachea: `ptrdiff_t` would also be correct here. This `offset` is guaranteed to be non-negative, so the signedness is irrelevant.

The `(char*)` is mandatory because pointer arithmetic on `void*` has unspecified behavior.

everythingfunctional: I understand that you can't do pointer math on a raw `void*`, but why `char` specifically?

bonachea: Any single-byte type would work; `char` is guaranteed to be a single byte. 

I usually use `uint8_t` (or `std::byte` in C++) for clarity/documentation in such circumstances, but they mean the same thing.

everythingfunctional: > I usually use `uint8_t` (or `std::byte` in C++) for clarity/documentation in such circumstances

Perhaps we should do so here too then. Or is it common to see something like `#define byte char`?

bonachea: >  is it common to see something like #define byte char?

It may be common, but I'd consider that a poor coding practice because: 
1) Never use a macro where a typedef would work, and 
2) `char` is "special" in a few other ways that IMO should not be obscured by macro/typedef.

I would not be opposed to `typedef uint8_t byte;` if we really want a shorthand, as long as such a declaration is never exported to Caffeine clients. I don't think we're planning to provide clients with any public header file in C for Caffeine, but I thought I'd mention it for completeness.

src/caffeine/caffeine.h
ktras: @bonachea I have added a note to the new README. The wording may be improved, or the section in which this note lives in the README may be changed.

src/caffeine/prif_queries_s.f90
ktras: Yes, that assert was wrong for 3 reasons. I have updated it and added tests as noted, I believe the logic is correct now, but of course this should be checked by others as well.

src/caffeine/caffeine.c
ktras: I have used the suggested typedef.

src/caffeine/allocation_m.f90
ktras: I agree that it seems like we should update the module structure so that we can keep the component `private`. I am happy to do it in a future PR after we discuss it further tomorrow and agree on a plan.

src/caffeine/caffeine.c
```diff
@@ -21,26 +25,25 @@ static gex_Rank_t rank, size;
   const int double_Complex_workaround =4100;
 #endif
 
-void caf_caffeinate(mspace* symmetric_heap, intptr_t* symmetric_heap_start, mspace* non_symmetric_heap, gex_TM_t* initial_team)
+void caf_caffeinate(mspace* symmetric_heap, intptr_t* symmetric_heap_start, mspace* non_symmetric_heap, gex_TM_t** initial_team)
```
bonachea, line 28: Why did you add a level of indirection in this declaration? As you just added to the README, you can not/must not modify a `bind(C)` function declaration in one language without updating the corresponding declaration in the other language, because no tool will help you notice the discrepancy.

However I believe this change is Just Plain Wrong.

src/caffeine/caffeine.c
```diff
@@ -56,12 +59,13 @@ void caf_caffeinate(mspace* symmetric_heap, intptr_t* symmetric_heap_start, mspa
...
   intptr_t non_symmetric_heap_start = *symmetric_heap_start + symmetric_heap_size;
 
-  if (caf_this_image(*initial_team) == 1) {
+  if (caf_this_image(myworldteam) == 1) {
     *symmetric_heap = create_mspace_with_base(*symmetric_heap_start, symmetric_heap_size, 0);
     mspace_set_footprint_limit(*symmetric_heap, symmetric_heap_size);
   }
   *non_symmetric_heap = create_mspace_with_base(non_symmetric_heap_start, non_symmetric_heap_size, 0);
   mspace_set_footprint_limit(*non_symmetric_heap, non_symmetric_heap_size);
+  *initial_team = &myworldteam;
```
bonachea, line 68: I suspect this incorrect assignment motivated the inadvisable change above. I believe this should instead have been:
```suggestion
  *initial_team = myworldteam;
```
I'd be surprised if this version in the current PR commit ran correctly at all...

ktras:
@bonachea @everythingfunctional I have implemented the requested change to make the global team a static variable in `caffeine.c` so we don't have to pass it back and forth for `caf_convert_base_addr()`. However, I don't think my changes are fully aligned with Brad's design to handle teams. In order to get the tests to pass and to have a working commit that everyone could look at to make this discussion easier, I had to return the collectives to using the static team variable, as they were in the past before Brad's adding of the coarray_handle on the Fortran side. This obviously isn't what we want in the long term, this is just temporary.

When trying to make my changes align with Brad's handling of teams, I had questions as to how we are passing around pointers between the C side and the Fortran side. In particular, I am unsure as to why, prior to my changes on this branch, the C definition of the argument for `initial_team` to `caf_caffeinate()` is `gex_TM_t* initial_team`, when the definition on the Fortran BIND(C) side is an `intent(out)` arg of type `c_ptr`. That would lead me expect a double pointer on the C side. And then similarly, I am not sure why the definition of the `team` arg for the collectives on the Fortran BIND(C) side is of type `c_ptr` but then on the C side, it is just `gex_TM_t team`. If I understand the answer to these questions, then I may be able to get closer to reconciling my changes with Brad's handling of teams for the collectives so far. We can also discuss this in our meeting tomorrow if it is easier to discuss live rather than online.

src/caffeine/caffeine.c
```diff
@@ -90,6 +93,15 @@ void caf_deallocate(mspace heap, void* mem)
   mspace_free(heap, mem);
 }
 
+// take address in a segment and convert to an address on given image
+intptr_t caf_convert_base_addr(void* addr, int image)
+{
+   ptrdiff_t offset = (byte*)addr - (byte*)gex_Segment_QueryAddr(mysegment);
+   void* segment_start_remote_image = NULL;
+   gex_Event_Wait(gex_EP_QueryBoundSegmentNB(myworldteam, image - 1, &segment_start_remote_image, NULL, NULL, 0));
+   return (intptr_t)((char *)segment_start_remote_image + offset);
```
bonachea, line 102: If we are adopting the convention of using `byte *` for pointer arithmetic in C, then please apply it consistently throughout this file.
```suggestion
   return (intptr_t)((byte *)segment_start_remote_image + offset);
```
There may be other occurences..

bonachea (changes requested):

src/caffeine/caffeine.c
ktras: I mention why I made this change in my comment I just barely posted above. Since `initial_team` is an `intent(out)` arg of type `c_ptr` in the BIND(C) declaration, I would expect it to be a double pointer on C side. I understand that I may be getting turned around with the passing of pointers back and forth between Fortran and C.

bonachea:
>  In order to get the tests to pass and to have a working commit that everyone could look at to make this discussion easier, I had to return the collectives to using the static team variable, as they were in the past before Brad's adding of the coarray_handle on the Fortran side. This obviously isn't what we want in the long term, this is just temporary.

The changes to `caf_caffeinate` are *mostly* correct. The changes to the collectives and image queries should be reverted immediately (in this PR). I suspect the error in the former motivated the latter.

src/caffeine/caffeine.c
bonachea: The `c_ptr` in the Fortran declaration is the `gex_TM_t` itself (which is a typedef to a C pointer). Thus the `c_ptr intent(out)` becomes `gex_TM_t *`

bonachea: Please also add a comment on the C definition of `caf_caffeinate` clarifying this detail to help the next person who misreads what's going on here.

src/caffeine/caffeine.c
ktras: @bonachea Thanks. After fixing this, the only other occurrence of a type cast with `char *` or `char*` is in the collectives, but no pointer math is being done on the resulting pointer.

src/caffeine/caffeine.c
ktras: Signature and definition has been updated and a note has been added.

src/caffeine/prif_queries_s.f90
```diff
@@ -13,6 +16,15 @@
   end procedure
 
   module procedure prif_base_pointer
+    integer(c_int) :: num_img
+
+    call prif_num_images(image_count=num_img)
+    call assert(image_num .ge. 0 .and. image_num .le. num_img, "prif_base_pointer: image_num not within valid range")
```
bonachea, line 22: Ok, so now the assertion appears to be logically correct.

Unfortunately as currently expressed the Fortran optimizer is unlikely be able to remove the associated overhead, because the call to `prif_num_images` currently includes an unconditional call to `caf_num_images() bind(C)` that is cross-module and cross-language, and this then invokes libgasnet (cross-library). This call chain is ridiculously over-complicated for the need at-hand (because the image data in the initial team is invariant for a given process over the lifetime of the job) and likely opaque to most analyses.

Ideas: 
1. `prif_init` should store such process info (image count and index in the initial team) as `prif` module variables in Fortran and just read them here, which should be alot more efficient and optimizable.
2. `caf_caffeinate` should also store this info as static variables in C-land so we also have very efficient access to that information in C.
3. We could optionally move or duplicate this assertion into `caf_convert_base_addr`, where we have hard guarantees that assertions compile away to nothing in production builds.

I'd really like to see options 1 and 2 deployed ASAP, as we'll eventually need similar checks elsewhere.

bonachea (changes requested):
Coming along nicely, made one additional request.

src/caffeine/prif_queries_s.f90
ktras: As discussed in our meeting today, we will hold off on this change until after other critical changes. Issue #59 tracks this. Eventual implementation will change `num_img` in the above line of code to `initial_team%num_images`.

src/caffeine/allocation_m.f90
bonachea: Added as issue #64

ktras: As discussed in our meeting and tracked in issue #64, this restructuring to keep components `private` will occur in a future PR.

bonachea (approved):
LGTM, thanks for entering issues for the outstanding discussions that will be resolved post-merge

Issue #81, opened 2024-03-20
Remove unnecessary argument definitions in collectives


rouson:
Until this issue is handled, any collective subroutine may crash.   For now, Caffeine will not provide error messages or statuses in the event of the failure of a collective subroutine:
- [x] Remove any handling of `errmsg` and `errmsg_alloc` from all collective subroutines
- [x] Set `stat=0` in all collective subroutines
- [x] Remove all references to unused variables in the C interfaces
- [x] Ensure all optional arguments in the prif collectives are checked if present before referencing them, unless passing them on to other optional arguments in a routine call

PR #82, opened 2024-03-20
Remove uses of unset and/or non-present variables in co_sum

rouson:


ktras (approved):
LGTM

bonachea:
For the record, this partially addresses issue #81, but only for `co_sum`

PR #86, opened 2024-03-26
Remove `stat` and `errmsg` args from `caf` collectives interfaces

### LINKED ISSUE #81: Remove unnecessary argument definitions in collectives

ktras:
Closes #81.

### DIFF
--- a/src/caffeine/caffeine.c
+++ b/src/caffeine/caffeine.c
@@ -179,9 +179,11 @@ static void set_stat_errmsg_or_abort(int* stat, char* errmsg, const int return_s
   }
 }
 
-void caf_co_max(CFI_cdesc_t* a_desc, int result_image, int* stat, char* errmsg, size_t num_elements, gex_TM_t team)
+void caf_co_max(CFI_cdesc_t* a_desc, int result_image, size_t num_elements, gex_TM_t team)
 {
   gex_DT_t a_type;
+  int* stat;
+  char* errmsg;
 
   switch (a_desc->type)
   {
--- a/src/caffeine/caffeine.c
+++ b/src/caffeine/caffeine.c
@@ -239,9 +239,11 @@ void caf_co_min(CFI_cdesc_t* a_desc, int result_image, size_t num_elements, gex_
   gex_Event_Wait(ev);
 }
 
-void caf_co_sum(CFI_cdesc_t* a_desc, int result_image, int* stat, char* errmsg, size_t num_elements, gex_TM_t team)
+void caf_co_sum(CFI_cdesc_t* a_desc, int result_image, size_t num_elements, gex_TM_t team)
 {
   gex_DT_t a_type;
+  int* stat = NULL;
+  char* errmsg = NULL;
 
   size_t c_sizeof_a = a_desc->elem_len;
 
@@ -267,8 +269,6 @@ void caf_co_sum(CFI_cdesc_t* a_desc, int result_image, int* stat, char* errmsg, 
     ev = gex_Coll_ReduceToAllNB(team,                 a_address, a_address, a_type, c_sizeof_a, num_elements, GEX_OP_ADD, NULL, NULL, 0);
   }
   gex_Event_Wait(ev);
-
-  if (stat != NULL) *stat = 0;
 }
 
 bool caf_same_cfi_type(CFI_cdesc_t* a_desc, CFI_cdesc_t* b_desc)
--- a/src/caffeine/caffeine.c
+++ b/src/caffeine/caffeine.c
@@ -250,7 +250,7 @@ void caf_co_sum(CFI_cdesc_t* a_desc, int result_image, size_t num_elements, gex_
     case float_Complex_workaround:  a_type = GEX_DT_FLT; num_elements *= 2; c_sizeof_a /= 2; break;
     case double_Complex_workaround: a_type = GEX_DT_DBL; num_elements *= 2; c_sizeof_a /= 2; break;
     default:
-      set_stat_errmsg_or_abort(NULL, NULL, UNRECOGNIZED_TYPE, "");
+      gasnett_fatalerror("Unrecognized type: %d", (int)a_desc->type);
   }
 
   char* a_address = (char*) a_desc->base_addr;
--- a/src/caffeine/caffeine.c
+++ b/src/caffeine/caffeine.c
@@ -182,8 +182,8 @@ static void set_stat_errmsg_or_abort(int* stat, char* errmsg, const int return_s
 void caf_co_max(CFI_cdesc_t* a_desc, int result_image, size_t num_elements, gex_TM_t team)
 {
   gex_DT_t a_type;
-  int* stat;
-  char* errmsg;
+  int* stat = NULL;
+  char* errmsg = NULL;
 
   switch (a_desc->type)
   {
@@ -207,8 +207,6 @@ void caf_co_max(CFI_cdesc_t* a_desc, int result_image, size_t num_elements, gex_
     ev = gex_Coll_ReduceToAllNB(team,                 a_address, a_address, a_type, c_sizeof_a, num_elements, GEX_OP_MAX, NULL, NULL, 0);
   }
   gex_Event_Wait(ev);
-
-  if (stat != NULL) *stat = 0;
 }
 
 void caf_co_min(CFI_cdesc_t* a_desc, int result_image, int* stat, char* errmsg, size_t num_elements, gex_TM_t team)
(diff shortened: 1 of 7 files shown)


src/caffeine/caffeine.c
```diff
@@ -183,9 +179,11 @@ static void set_stat_errmsg_or_abort(int* stat, char* errmsg, const int return_s
   }
 }
 
-void caf_co_max(CFI_cdesc_t* a_desc, int result_image, int* stat, char* errmsg, size_t num_elements, gex_TM_t team)
+void caf_co_max(CFI_cdesc_t* a_desc, int result_image, size_t num_elements, gex_TM_t team)
 {
   gex_DT_t a_type;
+  int* stat = NULL;
+  char* errmsg = NULL;
```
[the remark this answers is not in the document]
ktras: @everythingfunctional They are used in calls to the function `set_stat_errmsg_or_abort` if the type of the coarray is unsupported. Whether this is still necessary is another question that we should discuss. I meant to add this in the discussion of the PR but forget. (This is the same for the other local declarations added that you commented on).

src/caffeine/caffeine.c
```diff
@@ -239,13 +237,13 @@ void caf_co_min(CFI_cdesc_t* a_desc, int result_image, int* stat, char* errmsg,
...
-
-  if (stat != NULL) *stat = 0;
 }
 
-void caf_co_sum(CFI_cdesc_t* a_desc, int result_image, int* stat, char* errmsg, size_t num_elements, gex_TM_t team)
+void caf_co_sum(CFI_cdesc_t* a_desc, int result_image, size_t num_elements, gex_TM_t team)
 {
   gex_DT_t a_type;
+  int* stat = NULL;
+  char* errmsg = NULL;
```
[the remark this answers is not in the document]
ktras: See above reply to comment.

src/caffeine/caffeine.c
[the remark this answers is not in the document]
everythingfunctional: Was looking too fast and didn't see that. You could just pass `NULL` for now.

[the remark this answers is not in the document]
ktras: True. I have just made that change. Thanks.

src/caffeine/caffeine.c
```diff
@@ -194,7 +190,7 @@ void caf_co_max(CFI_cdesc_t* a_desc, int result_image, int* stat, char* errmsg,
     case CFI_type_float:            a_type = GEX_DT_FLT; break;
     case CFI_type_double:           a_type = GEX_DT_DBL; break;
     default:
-      set_stat_errmsg_or_abort(stat, errmsg, UNRECOGNIZED_TYPE, "");
+      set_stat_errmsg_or_abort(NULL, NULL, UNRECOGNIZED_TYPE, "");
```
bonachea, line 193: IIUC the call above is equivalent to `gasnett_fatalerror("")`, which is better than `abort()` but still not great: the message doesn't give a hint about what's wrong, due to the level of call indirection the failing line number won't even point to this line as the root cause, nor does it give a hint about what unrecognized type was passed.

```suggestion
      gasnett_fatalerror("Unrecognized type: %d", (int)a_desc->type);
```

src/caffeine/caffeine.c
```diff
@@ -224,7 +218,7 @@ void caf_co_min(CFI_cdesc_t* a_desc, int result_image, int* stat, char* errmsg,
...
```
bonachea, line 221:
```suggestion
      gasnett_fatalerror("Unrecognized type: %d", (int)a_desc->type);
```

src/caffeine/caffeine.c
```diff
@@ -258,7 +250,7 @@ void caf_co_sum(CFI_cdesc_t* a_desc, int result_image, int* stat, char* errmsg,
     case float_Complex_workaround:  a_type = GEX_DT_FLT; num_elements *= 2; c_sizeof_a /= 2; break;
     case double_Complex_workaround: a_type = GEX_DT_DBL; num_elements *= 2; c_sizeof_a /= 2; break;
     default:
-      set_stat_errmsg_or_abort(stat, errmsg, UNRECOGNIZED_TYPE, "");
+      set_stat_errmsg_or_abort(NULL, NULL, UNRECOGNIZED_TYPE, "");
```
bonachea, line 253:
```suggestion
      gasnett_fatalerror("Unrecognized type: %d", (int)a_desc->type);
```

src/caffeine/collective_subroutines/co_max_s.f90
```diff
@@ -1,32 +1,21 @@
...
-
       call caf_co_max( &
-          a, optional_value(result_image), stat_c_ptr, errmsg_c_ptr, int(product(shape(a)), c_size_t), current_team%gex_team)
-
-      call c_f_pointer(errmsg_c_ptr, errmsg_f_ptr) ! no need to do this for stat was passed by reference
-      errmsg = errmsg_f_ptr ! copy the output back & truncate the null terminator
-
+          a, optional_value(result_image), int(product(shape(a)), c_size_t), current_team%gex_team)
     else if (caf_is_f_string(a)) then
       call prif_co_reduce(a, c_funloc(reverse_alphabetize), optional_value(result_image), stat, errmsg)
```
bonachea, line 20: `errmsg_alloc` is missing here.

```suggestion
      call prif_co_reduce(a, c_funloc(reverse_alphabetize), optional_value(result_image), stat, errmsg, errmsg_alloc)
```

src/caffeine/collective_subroutines/co_min_s.f90
```diff
@@ -1,32 +1,21 @@
...
-
       call caf_co_min( &
-          a, optional_value(result_image), stat_c_ptr, errmsg_c_ptr, int(product(shape(a)), c_size_t), current_team%gex_team)
-
-      call c_f_pointer(errmsg_c_ptr, errmsg_f_ptr) ! no need to do this for stat was passed by reference
-      errmsg = errmsg_f_ptr ! copy the output back & truncate the null terminator
-
+          a, optional_value(result_image), int(product(shape(a)), c_size_t), current_team%gex_team)
     else if (caf_is_f_string(a)) then
       call prif_co_reduce(a, c_funloc(alphabetize), optional_value(result_image), stat, errmsg)
```
bonachea, line 20: `errmsg_alloc` is missing here.
```suggestion
      call prif_co_reduce(a, c_funloc(alphabetize), optional_value(result_image), stat, errmsg, errmsg_alloc)
```

bonachea (changes requested):
Made a few minor requests, otherwise LGTM

ktras:
@bonachea Thanks for the suggested changes. After making them, the function `set_stat_errmsg_or_abort` is never called. Is this a function we think will likely be useful in the future? Or should we delete the function? @rouson, @everythingfunctional, any thoughts?

everythingfunctional:
I would have changed `set_stat_errmsg_or_abort` to call `gasnett_fatalerror` instead of hard-coding that multiple places.

bonachea:
> I would have changed `set_stat_errmsg_or_abort` to call `gasnett_fatalerror` instead of hard-coding that multiple places.

@everythingfunctional what you suggest is essentially the "before" code. The "after" code I suggested with a direct call will generate a more useful error message.

bonachea:
> @bonachea Thanks for the suggested changes. After making them, the function `set_stat_errmsg_or_abort` is never called. Is this a function we think will likely be useful in the future? Or should we delete the function? @rouson, @everythingfunctional, any thoughts?

@ktras I'm in favor of removing the dead code. I don't expect we'll be supporting any non-fatal errors anytime soon other than allocation failure, and that doesn't use this code path.

everythingfunctional:
Ok, but then we should adjust the interface to `set_stat_errmsg_or_abort` to allow for passing the information/message we want, rather than getting rid of it. At some point we will want the logic it describes, and will be easier to make the change in one place instead of multiple.

That said, if you think

> I don't expect we'll be supporting any non-fatal errors anytime soon other than allocation failure, and that doesn't use this code path.

I'm fine with just delaying that refactoring for now.

bonachea (approved):

PR #95, opened 2024-03-28
Update docs for release 0.2.0

ktras:
- [x]  Update all instances of the copyright year embedded in:` LICENSE.txt`, `manifest/fpm.toml.template`
         - `LICENSE.txt` was already up to date
- [x]  Update all instances of the release package version number embedded in: `manifest/fpm.toml.template`, `install.sh`
- [x]  Update the author list embedded in: `manifest/fpm.toml.template`

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -29,10 +29,10 @@ list_prerequisites()
     echo ""
     echo "  GCC $GCC_VERSION"
     echo "  GASNet-EX $GASNET_VERSION"
-    echo "  fpm 0.5.0"
+    echo "  fpm 0.9.0"
     echo "  pkg-config 0.29.2"
-    echo "  realpath 9.0 (Homebrew coreutils 9.0)"
-    echo "  GNU Make 3.1 (Homebrew coreutils 9.0)"
+    echo "  realpath 9.4 (Homebrew coreutils 9.4)"
+    echo "  GNU Make 3.1 (Homebrew coreutils 9.4)"
     echo ""
     echo "In some cases, earlier versions might also work."
 }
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -29,10 +29,10 @@ list_prerequisites()
     echo ""
     echo "  GCC $GCC_VERSION"
     echo "  GASNet-EX $GASNET_VERSION"
-    echo "  fpm 0.5.0"
+    echo "  fpm 0.9.0"
     echo "  pkg-config 0.29.2"
-    echo "  realpath 9.0 (Homebrew coreutils 9.0)"
-    echo "  GNU Make 3.1 (Homebrew coreutils 9.0)"
+    echo "  realpath 9.4 (Homebrew coreutils 9.4)"
+    echo "  GNU Make 3.1 (Homebrew coreutils 9.4)"
```
bonachea, line 35: I thought we decided to drop version numbers from the Homebrew dependencies in this message?

IIUC We only actually require specific versions of GASNet and GCC.

bonachea (changes requested):

install.sh
ktras: As discussed in the meeting, particular version numbers are not needed, so the README changes by Damian mention these packages, but not particular version numbers. However, when I was updating this installer script, in the reading of the message at the beginning of this function, it says that this "prerequisite" list also informs what will be installed if the user doesn't already have them. So I checked the latest versions on Homebrew. But as I am writing this comment, I am realizing it would be better to just change that text that is shown to the user to say that the latest version on Homebrew will be installed. I'll come up with a new commit for feedback.

everythingfunctional (approved):

bonachea (approved):

PR #96, opened 2024-03-28
Create prefix

bonachea:
Teach install.sh to create --prefix directory if it doesn't already exist.   
This matches the default behavior of autotools and CMake that users generally expect.

Also, stop requiring gmkdir. 
Regular mkdir is sufficient for our purposes, and the install script
was already relying on its availability anyhow prior to installing   gmkdir.


ktras (approved):
LGTM

PR #99, opened 2024-03-28
Revert "Change installer to use specific gasnet version for the release"

ktras:
Revert temporary change to gasnet version in install script that is only needed for the release.


rouson (approved):
LGTM

PR #108, opened 2024-04-09
Fix calls in `caffeine.c`

### LINKED ISSUE #74: Remove warnings from `gcc` by fixing calls in `caffeine.c`
Remove warnings from `gcc` by fixing calls in `caffeine.c`

ktras:
Closes #74.

### DIFF
--- a/src/caffeine/caffeine.c
+++ b/src/caffeine/caffeine.c
@@ -35,6 +35,11 @@ typedef uint8_t byte;
   const int double_Complex_workaround =4100;
 #endif
 
+int caf_this_image(gex_TM_t team)
+{
+  return gex_TM_QueryRank(team) + 1;
+}
+
 // NOTE: gex_TM_T is a typedef to a C pointer, so the `gex_TM_t* initial_team` arg in the C signature matches the BIND(C) interface of an `intent(out)` arg of type `c_ptr` for the same argument
 void caf_caffeinate(mspace* symmetric_heap, intptr_t* symmetric_heap_start, mspace* non_symmetric_heap, gex_TM_t* initial_team)
 {
@@ -82,11 +87,6 @@ void caf_caffeinate(mspace* symmetric_heap, intptr_t* symmetric_heap_start, mspa
 void caf_decaffeinate(int exit_code)
 {
   gasnet_exit(exit_code);
-}
-
-int caf_this_image(gex_TM_t team)
-{
-  return gex_TM_QueryRank(team) + 1;
 }
 
 int caf_num_images(gex_TM_t team)
--- a/src/caffeine/caffeine.c
+++ b/src/caffeine/caffeine.c
@@ -71,10 +71,10 @@ void caf_caffeinate(mspace* symmetric_heap, intptr_t* symmetric_heap_start, mspa
   intptr_t non_symmetric_heap_start = *symmetric_heap_start + symmetric_heap_size;
 
   if (caf_this_image(myworldteam) == 1) {
-    *symmetric_heap = create_mspace_with_base(*symmetric_heap_start, symmetric_heap_size, 0);
+     *symmetric_heap = create_mspace_with_base((void*)*symmetric_heap_start, symmetric_heap_size, 0);
     mspace_set_footprint_limit(*symmetric_heap, symmetric_heap_size);
   }
-  *non_symmetric_heap = create_mspace_with_base(non_symmetric_heap_start, non_symmetric_heap_size, 0);
+  *non_symmetric_heap = create_mspace_with_base((void*)non_symmetric_heap_start, non_symmetric_heap_size, 0);
   mspace_set_footprint_limit(*non_symmetric_heap, non_symmetric_heap_size);
   *initial_team = myworldteam;
 }
--- a/src/caffeine/caffeine.c
+++ b/src/caffeine/caffeine.c
@@ -130,7 +130,7 @@ void caf_sync_all()
 }
 
 void caf_co_reduce(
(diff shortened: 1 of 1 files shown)


ktras:
The following two warnings still exist because when attempting to remove the warning by dereferencing the a`user_op` args to match what the functions seems to expect, then there is a runtime crash during the tests. These function calls involves macros expansions for both the function and for the argument type, so I may have misread what was needed to remove the warning.

```
././src/caffeine/caffeine.c:142:103: warning: passing argument 9 of 'gasnete_tm_reduce_nb' from incompatible pointer type [-Wincompatible-pointer-types]
  142 |       team, result_image-1, a_address, a_address, GEX_DT_USER, c_sizeof_a, num_elements, GEX_OP_USER, user_op, &c_sizeof_a, 0
      |                                                                                                       ^~~~~~~
      |                                                                                                       |
      |                                                                                                       void (**)(const void *, void *, size_t,  const void *) {aka void (**)(const void *, void *, long unsigned int,  const void *)}
/Users/kateras/.local/include/gasnet_coll.h:400:60: note: in definition of macro 'gex_Coll_ReduceToOneNB'
  400 |         gasnete_tm_reduce_nb(tm,root,dst,src,dt,dts,dtc,op,fn,cdata,flags,0 GASNETI_THREAD_GET)
      |                                                            ^~
/Users/kateras/.local/include/gasnet_coll.h:396:58: note: expected 'gex_Coll_ReduceFn_t' {aka 'void (*)(const void *, void *, long unsigned int,  const void *)'} but argument is of type 'void (**)(const void *, void *, size_t,  const void *)' {aka 'void (**)(const void *, void *, long unsigned int,  const void *)'}
  396 |                        gex_OP_t _op, gex_Coll_ReduceFn_t _user_op, void * _user_cdata,
      |                                      ~~~~~~~~~~~~~~~~~~~~^~~~~~~~
././src/caffeine/caffeine.c:146:103: warning: passing argument 8 of 'gasnete_tm_reduce_all_nb' from incompatible pointer type [-Wincompatible-pointer-types]
  146 |       team,                 a_address, a_address, GEX_DT_USER, c_sizeof_a, num_elements, GEX_OP_USER, user_op, &c_sizeof_a, 0
      |                                                                                                       ^~~~~~~
      |                                                                                                       |
      |                                                                                                       void (**)(const void *, void *, size_t,  const void *) {aka void (**)(const void *, void *, long unsigned int,  const void *)}
/Users/kateras/.local/include/gasnet_coll.h:413:59: note: in definition of macro 'gex_Coll_ReduceToAllNB'
  413 |         gasnete_tm_reduce_all_nb(tm,dst,src,dt,dts,dtc,op,fn,cdata,flags,0 GASNETI_THREAD_GET)
      |                                                           ^~
/Users/kateras/.local/include/gasnet_coll.h:409:59: note: expected 'gex_Coll_ReduceFn_t' {aka 'void (*)(const void *, void *, long unsigned int,  const void *)'} but argument is of type 'void (**)(const void *, void *, size_t,  const void *)' {aka 'void (**)(const void *, void *, long unsigned int,  const void *)'}
  409 |                         gex_OP_t _op, gex_Coll_ReduceFn_t _user_op, void * _user_cdata,
      |                                       ~~~~~~~~~~~~~~~~~~~~^~~~~~~~
```

bonachea:
> The following two warnings still exist because when attempting to remove the warnings by dereferencing the `user_op` args to match what the functions seems to expect, then there is a runtime crash during the tests.

I suspect you deployed the wrong fix. This is a warning about mismatched types in code that is already believed to operate correctly. In such a situation adding or removing a deference **operator** to assuage the typechecker is almost never the right fix, because it changes the semantics of the code that already runs correctly (to something that usually does not).

I believe the actual problem here is with the type declarations, specifically the C declaration of the `caf_co_reduce` function. `gex_Coll_ReduceFn_t` is already a pointer-to-function type, and thus compatible with the `c_funptr` provided from the Fortran side of the call. The problem is the extraneous `*` in the C declaration of the `user_op` formal argument to `caf_co_reduce`.

bonachea (changes requested):
Code changes so far look right.

I agree there are several problems revealed in issue #74 that remain unaddressed so far. I'd say it's your choice whether to merge this incremental improvement as-is, or attempt to fix the rest of the warnings here.

On a related note, I'd like to suggest/request a convention where commits which fix or address issues include the issue number **in the commit message**. Right now the issue number appears only in the PR description, but years from now it will be much easier to understand the commit in isolation if the issue number also appears in the commit message.

src/caffeine/caffeine.c
```diff
@@ -71,10 +76,10 @@ void caf_caffeinate(mspace* symmetric_heap, intptr_t* symmetric_heap_start, mspa
   intptr_t non_symmetric_heap_start = *symmetric_heap_start + symmetric_heap_size;
 
   if (caf_this_image(myworldteam) == 1) {
-    *symmetric_heap = create_mspace_with_base(*symmetric_heap_start, symmetric_heap_size, 0);
+     *symmetric_heap = create_mspace_with_base((void*)*symmetric_heap_start, symmetric_heap_size, 0);
```
everythingfunctional, line 79: Small nitpick to delete the extraneous whitespace.

everythingfunctional (approved):
Looks fine to me, but I'll defer to Dan on the intricacies of C type-casting.

src/caffeine/caffeine.c
ktras: Thanks for the catch. Fixed.

bonachea (approved):

PR #109, opened 2024-04-09
Remove repetition in `install.sh` by using heredoc

### LINKED ISSUE #104: Remove repetition in `install.sh` by using heredoc
https://linuxize.com/post/bash-heredoc/

ArthurPV:
Close #104

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -4,18 +4,20 @@ set -e # exit on error
 
 print_usage_info()
 {
-    echo "Caffeine Installation Script"
-    echo ""
-    echo "USAGE:"
-    echo "./install.sh [--help | [--prefix=PREFIX]"
-    echo ""
-    echo " --help             Display this help text"
-    echo " --prefix=PREFIX    Install library into 'PREFIX' directory"
-    echo " --prereqs          Display a list of prerequisite software."
-    echo "                    Default prefix='\$HOME/.local/bin'"
-    echo ""
-    echo "For a non-interactive build with the 'yes' utility installed, execute"
-    echo "yes | ./install.sh"
+    cat << EOF
+Caffeine Installation Script
+
+USAGE:
+./install.sh [--help | [--prefix=PREFIX]
+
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -95,31 +99,34 @@ fi
...
+EOF
 }
 
 ask_permission_to_install_homebrew_package()
 {
   echo ""
   if [ ! -z ${2+x} ]; then
-    echo "Homebrew installs $1 collectively in one package named '$2'."
-    echo ""
+    echo -e "Homebrew installs $1 collectively in one package named '$2'.\n"
```
bonachea, line 129: `echo -e` is non-POSIX and not available when using `/bin/sh` on many versions of macos, which currently drives this script. 

This patch hunk needs to be removed.

install.sh
```diff
@@ -95,31 +99,34 @@ fi
...
 ask_permission_to_use_homebrew()
 {
-  echo ""
-  echo "Either one or more of the environment variables FC, CC, and CXX are unset or"
-  echo "one or more of the following packages are not in the PATH: pkg-config, realpath, make, fpm."
-  echo "If you grant permission to install prerequisites, you will be prompted before each installation." 
-  echo ""
-  echo "Press 'Enter' to choose the square-bracketed default answer:"
-  printf "Is it ok to use Homebrew to install prerequisite packages? [yes] "
+  cat << EOF | head -c -1
```
bonachea, line 102: These `head -c -1` commands are non-POSIX and generate an error on macos

bonachea (changes requested):
@ArthurPV Thanks for the contribution!

Unfortunately these changes include several defects that make it unusable on macOS, at least in its current form. We need the installer to work portably on recent versions of both Linux and macOS.

install.sh
ArthurPV: Okay I see, I'll make the change.

bonachea (approved):
LGTM!

Issue #119, opened 2024-07-17
Update `prif` module based on Version 0.4 changes


ktras:
- [x] Add `PRIF_VERSION_MAJOR`, `PRIF_VERSION_MINOR`
- [x] Add `prif_event_type`
- [x] Remove `prif_base_pointer` from `prif` and rename as `caf_base_pointer`
- [x] `prif_init` - **no change needed to interface**
- [x] `prif_stop` - **no change needed to interface**
- [x] `prif_error_stop` - **no change needed to interface**
- [x] `prif_fail_image` - **no change needed to interface**
- [x] `prif_num_images` - **implemented**
- [x] `prif_num_images_with_team` - **NOT implemented**
- [x] `prif_num_images_with_team_number` - **NOT implemented**
- [x] `prif_this_image_no_coarray` - **implemented** (optional arg `team` not handled)
- [x] `prif_this_image_with_coarray` - **NOT implemented**
- [x] `prif_this_image_with_dim` - **NOT implemented**
- [x] `prif_failed_images` - **no change needed to interface, NOT implemented**
- [x] `prif_stopped_images` - **no change needed to interface, NOT implemented**
- [x] `prif_image_status` - **no change needed to interface? have added procedure keywords compared to PRIF spec**
- [x] `prif_allocate_coarray` - **implemented**
- [x] `prif_allocate` - **implemented**
- [x] `prif_deallocate_coarray` - **implemented**
- [x] `prif_deallocate` - **implemented**
- [x] `prif_alias_create` - **NOT implemented**
- [x] `prif_alias_destroy` - **no change needed to interface, NOT implemented**
- [x] `prif_set_context_data` - **no change needed to interface, NOT implemented**
- [x] `prif_get_context_data` - **no change needed to interface, NOT implemented**
- [x] `prif_size_bytes` - **no change needed to interface, NOT implemented**
- [x] `prif_lcobound_with_dim` - **removed generic interface, specific interface unchanged, NOT implemented**
- [x] `prif_lcobound_no_dim` - **removed generic interface, specific interface unchanged, NOT implemented**
- [x] `prif_ucobound_with_dim`  - **removed generic interface, specific interface unchanged, NOT implemented**
- [x] `prif_ucobound_no_dim` - **removed generic interface, specific interface unchanged, NOT implemented**
- [x] `prif_coshape` - **no change needed to interface, NOT implemented**
- [x] `prif_image_index` - **implemented**
[...]

PR #120, opened 2024-07-22
Update PRIF to version 0.4 for all interface changes except for puts and gets

ktras:
Part of issue #119.

### DIFF
--- a/src/prif.f90
+++ b/src/prif.f90
@@ -459,11 +459,32 @@ module prif
       character(len=:), intent(inout), allocatable, optional :: errmsg_alloc
     end subroutine
 
+    module subroutine prif_lock(image_num, coarray_handle, offset, acquired_lock, stat, errmsg, errmsg_alloc)
+      implicit none
+      integer(c_int), intent(in) :: image_num
+      type(prif_coarray_handle), intent(in) :: coarray_handle
+      integer(c_size_t), intent(in) :: offset
+      logical(c_bool), intent(out), optional :: acquired_lock
+      integer(c_int), intent(out), optional :: stat
+      character(len=*), intent(inout), optional :: errmsg
+      character(len=:), intent(inout), allocatable, optional :: errmsg_alloc
+    end subroutine
+
     module subroutine prif_lock_indirect(image_num, lock_var_ptr, acquired_lock, stat, errmsg, errmsg_alloc)
       implicit none
       integer(c_int), intent(in) :: image_num
       integer(c_intptr_t), intent(in) :: lock_var_ptr
       logical(c_bool), intent(out), optional :: acquired_lock
+      integer(c_int), intent(out), optional :: stat
+      character(len=*), intent(inout), optional :: errmsg
+      character(len=:), intent(inout), allocatable, optional :: errmsg_alloc
+    end subroutine
+
+    module subroutine prif_unlock(image_num, coarray_handle, offset, stat, errmsg, errmsg_alloc)
+      implicit none
+      integer(c_int), intent(in) :: image_num
+      type(prif_coarray_handle), intent(in) :: coarray_handle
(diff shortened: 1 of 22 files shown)


src/prif.f90
bonachea, line 1: Please remove `use iso_fortran_env`

bonachea (changes requested):
Made one minor request, otherwise LGTM based on a casual skim and a few spot-checks. I did not attempt to exhaustively compare every interface declaration to the spec.

src/prif.f90
ktras: Nice, nice catch! Thanks! Done.

bonachea (approved):
LGTM!

src/prif.f90
```diff
@@ -49,55 +57,8 @@ module prif
...
-      implicit none
-      type(prif_coarray_handle), intent(in) :: coarray_handle
-      integer(c_int), intent(in) :: dim
-      type(prif_team_type), intent(in), optional :: team
-      integer(c_intmax_t), intent(out) :: cosubscript
-    end subroutine
-
-  end interface
+  integer(c_int), parameter, public :: PRIF_ATOMIC_INT_KIND = selected_int_kind(18)
+  integer(c_int), parameter, public :: PRIF_ATOMIC_LOGICAL_KIND = PRIF_ATOMIC_INT_KIND
```
everythingfunctional, line 61: The use of an integer kind as a logical kind is definitely not portable, but I kind of think we should use the C kinds here (corresponding to what GASNet uses for atomic operations).

everythingfunctional (commented):
Looks good otherwise. Thanks Kate.

src/prif.f90
everythingfunctional: @bonachea , says use `selected_logical_kind(32)`, and I agree.

bonachea:
```suggestion
  integer(c_int), parameter, public :: PRIF_ATOMIC_INT_KIND = selected_int_kind(18)
  integer(c_int), parameter, public :: PRIF_ATOMIC_LOGICAL_KIND = selected_logical_kind(32)
```

everythingfunctional: I don't have any idea what their rationale would have been for specifying it that way, but I doubt I would have agreed with it.

bonachea: IIUC @everythingfunctional 's motivation for this change is because int kind parameters are not guaranteed to be meaningful logical kind parameters. So the old code was very non-portable.

However I think it's worth noting that even this new implementation assumes the compiler HAS a logical type of at least 32 bits. Hopefully flang provides that?

IIUC a Fortran compiler could(?) be compliant providing only an 8-bit logical kind, and such a compiler would return `-1` above, rendering Caffeine erroneous. If we discover there is such a compiler of interest out there, we'd need considerably more effort to support logical atomics, and we'd probably need some configure-time knobs to activate that support.

everythingfunctional: > IIUC @everythingfunctional 's motivation for this change is because int kind parameters are not guaranteed to be meaningful logical kind parameters. So the old code was very non-portable.

Correct.

> However I think it's worth noting that even this new implementation assumes the compiler HAS a logical type of at least 32 bits.

> IIUC a Fortran compiler could(?) be compliant providing only an 8-bit logical kind, and such a compiler would return `-1` above, rendering Caffeine erroneous.

This is true, but seems unlikely to occur in practice, and we can probably cross that bridge when (if) we come to it.

ktras: Just to summarize conversations we had elsewhere related to this: gfortran doesn't support `selected_logical_kind`. As such, this code suggestion has been reverted and then a comment documenting this has been added to the code.

bonachea (approved):

everythingfunctional (approved):

PR #121, opened 2024-07-24
Update puts and gets for prif v0.4

### LINKED ISSUE #67: Optimize/Inline `as_int` and `as_ptr`
Based on my explorations it should be "valid" to implement these as `transfer(p, i)` and `transfer(i, p)` respectively and not need to call C functions. That said we should potentially safeguard that with a check that `sizeof(i) == sizeof(p)`.

everythingfunctional:
This changes the names and interfaces of the existing put/get routines to match PRIF 0.4, removes `prif_base_pointer` from the public interface, and fixes #67


src/caffeine/prif_private_s.f90
```diff
@@ -185,24 +185,45 @@ pure function caf_elem_len(a) result(a_elem_len) bind(C)
...
+
+    integer(c_int) :: num_img
+
+    call prif_num_images(num_images=num_img)
+    call assert(image_num .ge. 0 .and. image_num .le. num_img, "caf_base_pointer: image_num not within valid range")
+    if (image_num .eq. 0) then
+      ptr = 0
+    else
+      ptr = caf_convert_base_addr(coarray_handle%info%coarray_data, image_num)
+    end if
```
bonachea, line 224: I suspect this branch is now irrelevant.

IIRC `prif_base_pointer()` was specified to accept `image_num==0` and return a null pointer specifically to allow the client to pass the (possibly empty) result of `image_index()` directly to `prif_base_pointer()` before checking the result for validity. 

However the client no longer calls `prif_base_pointer()`, and passing `image_num==0` to the communication routines is an error. As such I think this branch should be removed and the assertion above adjusted to exclude zero as a valid input.

```suggestion
    call assert(image_num .gt. 0 .and. image_num .le. num_img, "caf_base_pointer: image_num not within valid range")
    ptr = caf_convert_base_addr(coarray_handle%info%coarray_data, image_num)
```

src/caffeine/prif_private_s.f90
```diff
@@ -185,24 +185,45 @@ pure function caf_elem_len(a) result(a_elem_len) bind(C)
...
-       import c_ptr, c_intptr_t
-       integer(c_intptr_t), intent(in), value :: i
-       type(c_ptr) :: caf_as_c_ptr
-     end function
-
   end interface
 
 contains
 
+  pure function caf_as_int(ptr)
```
bonachea, line 192: IIUC Caffeine currently follows the convention that `caf_` names are only used for `BIND(C)` procedures, so it's easy to see at the call-site when we are making a cross-language call (which is probably not subject to normal inlining, and thus potentially a more expensive call). Private Caffeine procedures implemented in Fortran have no prefix (e.g. `assert`, `unimplemented`, `optional_value`), because they are already hidden by the module privacy.

This seems like a useful convention to preserve. Assuming you agree, let's rename these three converted procedures appropriately and update the coding convention in <docs/README-maintainers.md>

src/caffeine/prif_private_s.f90
```diff
@@ -185,24 +185,45 @@ pure function caf_elem_len(a) result(a_elem_len) bind(C)
...
+  end function
+
+  subroutine caf_base_pointer(coarray_handle, image_num, ptr)
+    type(prif_coarray_handle), intent(in) :: coarray_handle
+    integer(c_int), intent(in) :: image_num
+    integer(c_intptr_t), intent(out) :: ptr
+
+    integer(c_int) :: num_img
+
+    call prif_num_images(num_images=num_img)
```
bonachea, line 218: This high-impact subroutine currently makes two cross-language calls where only one is required.

Suggest a todo to fix this later (once team support is merged)
```suggestion
    ! TODO TEAMS: either move the assertion below into `caf_convert_base_addr()`
    ! (avoiding the need to call prif_num_images here to fetch initial team size)
    ! or cache the initial team size in a private module variable so we can just access it
    call prif_num_images(num_images=num_img)
```

bonachea (changes requested):
I know this is still draft, but I added a few minor requests

src/caffeine/prif_private_s.f90
ktras: I agree. I think that is a good convention to have. If @everythingfunctional agrees with that convention, I am happy to update the convention in that doc on a separate branch and PR. Or it can become part of this PR, whichever is best.

To make sure I follow the suggested convention, it would go something like this right?:

_Procedures with the prefix `prif_` are public procedures that are part of the PRIF specification.
Procedures with the prefix `caf_` are internal Caffeine procedures that are `BIND(C)` and whose implementation is in C.
Procedures with no prefix are internal private Caffeine procedures whose implementation is in Fortran._

bonachea:
When convenient, please rebase onto main to hide the common commits from PR #120

src/caffeine/prif_private_s.f90
everythingfunctional: That convention seems reasonable, so I've done as @bonachea suggested.

src/caffeine/prif_private_s.f90
everythingfunctional: Agreed, I've now done it this way.

bonachea (commented):
I see this PR has been promoted out of "Draft" status.

The changes so far LGTM, although it's still a ways off from providing interface declarations for all the variants of put/get procedures required by PRIF 0.4 (listed in issue #119).

Is the intent to merge this as-is and start a new PR for the remaining interface declarations?

To be clear, actually fully implementing all the different variants of put/get is non-trivial and probably should not be a near-term goal. However I thought our near-term goal was to complete the declarations in prif.f90 (making that file statically 0.4 compliant), even though many of the procedures will simply `call unimplemented()` at runtime for now (and thus won't have any tests yet).

Thoughts?

everythingfunctional:
> Is the intent to merge this as-is and start a new PR for the remaining interface declarations?

That was my thought, but I could be convinced to add the other interfaces as part of this PR if you think it would be better.

bonachea (approved):
I agree with Brad's plan to merge this now as-is, and update issue #119 to indicate the remaining known divergences from PRIF 0.4. 

A subsequent PR (hopefully soon) will add interface declarations for the remaining put/get procedures with `unimplemented()` implementations (for now).

docs/README-maintainers.md
```diff
@@ -21,7 +21,10 @@ Conventions for code and commits in Caffeine
 -------------
 [TODO: Improve wording of below bullets]
 * Procedures with `prif_` prefix are public facing procedures
-* Procedures with `caf_` prefix are private procedures internal to Caffeine
+* Procedures with `caf_` prefix are private procedures,
+  internal to Caffeine, which are implemented in C
+* Procedures with neither of the above two prefixes are private procedures,
+  internal to Caffeine. which are implemented in Fortran
```
ktras, line 27:
```suggestion
  internal to Caffeine, which are implemented in Fortran
```

ktras (changes requested):
LGTM with one small code suggestion

ktras (approved):
LGTM

Issue #123, opened 2024-07-25
Add level of indirection to Caffeine implementation of `prif_team_type`


bonachea:
Currently in prif.f90:

```Fortran
  type, public :: prif_team_type
    private
    type(c_ptr) :: gex_team
    type(c_ptr) :: heap_mspace
    integer(c_intptr_t) :: heap_start
    type(prif_team_type), pointer :: parent_team
    type(handle_data), pointer :: coarrays
  end type
```

This "unboxed" implementation of `prif_team_type` contains several mutable fields, and the [WIP teams branch](https://github.com/BerkeleyLab/caffeine/compare/main...teams-support) adds even more.

As such, I believe this is an unsuitable representation for `prif_team_type`: because it fails to provide the handle-like semantics of `ISO_FORTRAN_ENV::TEAM_TYPE`. IOW the client is free to make copies of `TEAM_TYPE`/`prif_team_type` variables, and all copies of the same `prif_team_type` value need to consistently reference the same abstract team, even after operations like `CHANGE TEAM` and `prif_allocate_coarray` that mutate the team's internal state.

The solution is to add a level of indirection, similar to what we already have `prif_coarray_handle` (for some of the same reasons). Eg:

```Fortran
  type, public :: prif_team_type
    private
    type(team_data), pointer :: info
  end type

...

  type, private :: team_data
    private
    type(c_ptr) :: gex_team
    type(c_ptr) :: heap_mspace
    integer(c_intptr_t) :: heap_start
    type(team_data), pointer :: parent_team
    type(handle_data), pointer :: coarrays
  end type
```

Along with updates to all the code that accesses `prif_team_type` components.

bonachea:
Closely related: #59

bonachea:
Resolved in #127

PR #126, opened 2024-08-06
Implement teams support

everythingfunctional:
This implements our basic strategy for supporting teams.

### DIFF
--- a/src/caffeine/allocation_s.f90
+++ b/src/caffeine/allocation_s.f90
@@ -7,7 +7,6 @@ submodule(prif:prif_private_s) allocation_s
       c_f_procpointer, &
       c_loc, &
       c_associated, &
-      c_null_ptr, &
       c_null_funptr
 
   implicit none
@@ -55,7 +54,7 @@ contains
     coarray_handle%info%final_func = final_func
     coarray_handle%info%lcobounds(1:size(lcobounds)) = lcobounds
     coarray_handle%info%ucobounds(1:size(ucobounds)) = ucobounds
-    call add_to_team_list(current_team, coarray_handle)
+    call add_to_team_list(coarray_handle)
 
     allocated_memory = coarray_handle%info%coarray_data
     if (caf_have_child_teams()) then
@@ -140,20 +139,14 @@ contains
     call caf_deallocate(non_symmetric_heap_mspace, mem)
   end procedure
 
-  subroutine add_to_team_list(current_team, coarray_handle)
-    type(prif_team_type), intent(inout) :: current_team
+  subroutine add_to_team_list(coarray_handle)
     type(prif_coarray_handle), intent(inout) :: coarray_handle
 
     if (associated(current_team%info%coarrays)) then
       current_team%info%coarrays%previous_handle = c_loc(coarray_handle%info)
       coarray_handle%info%next_handle = c_loc(current_team%info%coarrays)
-      coarray_handle%info%previous_handle = c_null_ptr
-      current_team%info%coarrays => coarray_handle%info
-    else
-      current_team%info%coarrays => coarray_handle%info
-      coarray_handle%info%next_handle = c_null_ptr
-      coarray_handle%info%previous_handle = c_null_ptr
     end if
+    current_team%info%coarrays => coarray_handle%info
   end subroutine
 
   subroutine remove_from_team_list(coarray_handle)
@@ -161,9 +154,17 @@ contains
 
     type(handle_data), pointer :: tmp_data
 
+    if (&
+        .not.c_associated(coarray_handle%info%previous_handle) &
+        .and. .not.c_associated(coarray_handle%info%next_handle)) then
+      nullify(current_team%info%coarrays)
+      return
+    end if
     if (c_associated(coarray_handle%info%previous_handle)) then
       call c_f_pointer(coarray_handle%info%previous_handle, tmp_data)
       tmp_data%next_handle = coarray_handle%info%next_handle
+    else
+      call c_f_pointer(coarray_handle%info%next_handle, current_team%info%coarrays)
     end if
     if (c_associated(coarray_handle%info%next_handle)) then
       call c_f_pointer(coarray_handle%info%next_handle, tmp_data)
--- a/src/caffeine/allocation_s.f90
+++ b/src/caffeine/allocation_s.f90
@@ -140,7 +140,7 @@ contains
   end procedure
 
   subroutine add_to_team_list(coarray_handle)
-    type(prif_coarray_handle), intent(inout) :: coarray_handle
+    type(prif_coarray_handle), intent(in) :: coarray_handle
 
     if (associated(current_team%info%coarrays)) then
       current_team%info%coarrays%previous_handle = c_loc(coarray_handle%info)
--- a/src/caffeine/allocation_s.f90
+++ b/src/caffeine/allocation_s.f90
@@ -28,6 +28,12 @@ contains
     coarray_size = product(ubounds-lbounds+1)*element_size
 
     me = caf_this_image(current_team%info%gex_team)
+    if (caf_have_child_teams()) then
+      ! Free the child team space to make sure we have space to allocate the coarray
+      if (me == 1) then
+        call caf_deallocate(current_team%info%heap_mspace, current_team%info%child_heap_info%allocated_memory)
+      end if
+    end if
     if (me == 1) then
       handle_size = c_sizeof(unused)
       total_size = handle_size + coarray_size
@@ -52,6 +58,9 @@ contains
     call add_to_team_list(current_team, coarray_handle)
 
     allocated_memory = coarray_handle%info%coarray_data
+    if (caf_have_child_teams()) then
+      call caf_establish_child_heap
+    end if
   end procedure
 
   module procedure prif_allocate
@@ -118,6 +127,13 @@ contains
         call caf_deallocate(current_team%info%heap_mspace, c_loc(coarray_handles(i)%info))
     end do
     if (present(stat)) stat = 0
+    if (caf_have_child_teams()) then
+      ! reclaim any free space possible for the child teams to use
+      if (caf_this_image(current_team%info%gex_team) == 1) then
+        call caf_deallocate(current_team%info%heap_mspace, current_team%info%child_heap_info%allocated_memory)
+      end if
+      call caf_establish_child_heap
+    end if
   end procedure
 
   module procedure prif_deallocate
--- a/src/caffeine/prif_private_s.f90
+++ b/src/caffeine/prif_private_s.f90
@@ -191,6 +191,15 @@ submodule(prif) prif_private_s
        integer(c_size_t), target :: a_elem_len
      end function
 
+     subroutine caf_form_team(current_team, new_team, team_number, new_index) bind(C)
+      !! void caf_form_team(gex_TM_t* current_team, gex_TM_t* new_team, intmax_t team_number, int new_index);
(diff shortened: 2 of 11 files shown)


src/prif.f90
```diff
@@ -53,8 +53,11 @@ module prif
     type(c_ptr) :: gex_team
     type(c_ptr) :: heap_mspace
     integer(c_intptr_t) :: heap_start
+    integer(c_size_t) :: heap_size
     type(prif_team_type), pointer :: parent_team
     type(handle_data), pointer :: coarrays
+    type(prif_coarray_handle) :: child_team_handle
+    type(child_team_info), pointer :: child_heap_info ! offset and size
```
bonachea, line 60: As explained in #123, I believe this representation of `prif_team_type` is a fundamentally incorrect implementation of `ISO_FORTRAN_ENV::TEAM_TYPE`, because it includes mutable team state unboxed inside the client-owned object. I expect this can lead to breakage as soon as the client makes a copy of a `TEAM_TYPE` aka `prif_team_type` value and then uses one copy to `CHANGE TEAM`.

I think we should go ahead and add the missing level of indirection requested in #123 ASAP, because the current declarations in prif.f90 mislead potential PRIF implementers into believing the current declarations represent a valid implementation strategy.

bonachea (changes requested):
I realize this is still draft, but I went ahead and added one high-level comment/request

src/prif.f90
everythingfunctional: I made this chance as part of a separate PR to get it in sooner, and then rebased this branch on to that.

src/caffeine/teams_s.f90
```diff
@@ -1,20 +1,70 @@
...
+    block
+      integer(c_int) :: new_index_
+      if (present(new_index)) then
+        new_index_ = new_index
+      else
+        new_index_ = 1
+      end if
+
+      team%info%parent_team => current_team%info
+      call caf_form_team(current_team%info%gex_team, team%info%gex_team, team_number, new_index)
```
bonachea, line 66: Pretty sure you meant to pass `new_index_` here

```suggestion
      call caf_form_team(current_team%info%gex_team, team%info%gex_team, team_number, new_index_)
```

src/caffeine/teams_s.f90
```diff
@@ -1,20 +1,70 @@
...
+    end if
   end procedure
 
   module procedure prif_end_team
-    call unimplemented("prif_end_team")
+    type(prif_coarray_handle), allocatable :: teams_coarrays(:)
+    integer :: num_coarrays_in_team, i
+    type(handle_data), pointer :: tmp_data
+
+    ! deallocate the teams coarrays
```
bonachea, line 29: The two-pass algorithm with a heap temporary below could instead have been written (more simply) as a one-pass that deallocates each coarray individually. 

I assume you wrote it this way in order to batch together multiple deallocations into a single `prif_deallocate_coarray()` call?

This seems like the right choice, but it's non-obvious at first glance. Request a comment to explain this, eg:

```suggestion
    ! deallocate the teams coarrays
    ! Currently we work to batch together all the deallocations into a single call
    ! to prif_deallocate_coarray(), in the hope it can amortize some costs
```

src/caffeine/teams_s.f90
```diff
@@ -1,20 +1,70 @@
...
+      call c_f_pointer(tmp_data%next_handle, tmp_data)
+    end do
+    if (num_coarrays_in_team > 0) then
+      allocate(teams_coarrays(num_coarrays_in_team))
+      tmp_data => current_team%info%coarrays
+      do i = 1, num_coarrays_in_team
+        teams_coarrays(i)%info => tmp_data
+        call c_f_pointer(tmp_data%next_handle, tmp_data)
+      end do
+      call prif_deallocate_coarray(teams_coarrays, stat, errmsg, errmsg_alloc)
```
bonachea, line 43: IIUC `current_team%info%coarrays` is a dangling pointer now and needs to be nullified to prevent an error after the next change team to this team.

```suggestion
      call prif_deallocate_coarray(teams_coarrays, stat, errmsg, errmsg_alloc)
      nullify(current_team%info%coarrays)
```

src/caffeine/prif_private_s.f90
```diff
@@ -238,4 +269,21 @@ pure function optional_value(var) result(c_val)
...
+    if (caf_this_image(current_team%info%gex_team) == 1) then
+      call caf_allocate_remaining( &
+          current_team%info%heap_mspace, &
+          current_team%info%child_heap_info%allocated_memory, &
+          current_team%info%child_heap_info%size)
+      current_team%info%child_heap_info%offset = &
+          as_int(current_team%info%child_heap_info%allocated_memory) - current_team%info%heap_start
+    end if
+    call prif_co_broadcast(current_team%info%child_heap_info%offset, 1)
+    call prif_co_broadcast(current_team%info%child_heap_info%size, 1)
```
bonachea, line 282: Collective operations have a high latency cost, so it will be nearly twice as fast to broadcast the entire `child_team_info` in a single broadcast operation, even if you only need two of the fields.

```suggestion
    call prif_co_broadcast(current_team%info%child_heap_info, 1)
```

This optimization is especially relevant since we currently call this subroutine during every change team or coarray allocation when the current team has children.

bonachea (changes requested):
Coming along nicely!

Added a few more suggestions

src/caffeine/teams_s.f90
everythingfunctional: Good catch. I had to find it the hard way.

src/caffeine/teams_s.f90
everythingfunctional: You're right, I like your suggested comment so I added it.

src/caffeine/teams_s.f90
everythingfunctional: Another good catch. I now do so.

src/caffeine/prif_private_s.f90
everythingfunctional: Good point.

everythingfunctional:
I think this is the design/implementation we want now, but I'm not understanding something that is wrong with the linked list of coarrays on images other than 1 in a team. It's not treating pointers that have been set to null as being null, but only on images other than 1.

bonachea (changes requested):
New test looks good, and we seem to be getting closer!

I didn't notice a defect likely to explain exactly the runtime crash you report. However I did notice and comment on a few important runtime checks that are currently missing and might potentially explain other runtime crashes.

Also added a few other misc minor requests.

test/caf_teams_test.f90
```diff
@@ -0,0 +1,62 @@
...
+        type(prif_coarray_handle) :: coarrays(num_coarrays)
+        type(c_ptr) :: allocated_memory
+        type(prif_team_type) :: team
+
+        call prif_this_image_no_coarray(this_image=me)
+        which_team = merge(1_c_intmax_t, 2_c_intmax_t, mod(me, 2) == 0)
+        element_size = int(storage_size(dummy_element)/8, c_size_t)
+        call prif_form_team(team_number = which_team, team = team)
+        call prif_change_team(team)
+            call prif_num_images(num_images=num_imgs)
```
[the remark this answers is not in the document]
everythingfunctional: Seems reasonable. I've added this to the test.

src/caffeine/teams_s.f90
```diff
@@ -1,20 +1,75 @@
...
+    if (caf_have_child_teams()) then ! need to establish heap for child teams
+      call caf_establish_child_heap
+    end if
   end procedure
 
   module procedure prif_end_team
-    call unimplemented("prif_end_team")
+    type(prif_coarray_handle), allocatable :: teams_coarrays(:)
+    integer :: num_coarrays_in_team, i
+    type(c_ptr) :: tmp_c_ptr
```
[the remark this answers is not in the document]
everythingfunctional: Oops. Good catch.

test/caf_teams_test.f90
```diff
@@ -0,0 +1,62 @@
...
+contains
+    function test_caf_teams() result(tests)
+        type(test_item_t) :: tests
+
+        tests = describe( &
+            "Teams", &
+            [ it("can be created, changed to, and allocate coarrays", check_teams) &
+            ])
+    end function
+
```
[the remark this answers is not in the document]
everythingfunctional: Ideally tests for different features are separate tests. Admittedly such a test involves the interaction of different features. It may be worth a brief conversation on our testing strategy/organisation.

src/caffeine/allocation_s.f90
```diff
@@ -118,36 +126,45 @@
...
   end procedure
 
   module procedure prif_deallocate
     call caf_deallocate(non_symmetric_heap_mspace, mem)
   end procedure
 
-  subroutine add_to_team_list(current_team, coarray_handle)
-    type(prif_team_type), intent(inout) :: current_team
+  subroutine add_to_team_list(coarray_handle)
     type(prif_coarray_handle), intent(inout) :: coarray_handle
```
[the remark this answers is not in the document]
everythingfunctional: Yeah, I keep struggling with this conceptually. Sure the pointer isn't changing, but the data it points to is, so my intuition defaults to `inout`. You're right though, since the pointer isn't changing it can be `intent(in)`.

src/caffeine/teams_s.f90
```diff
@@ -1,20 +1,75 @@
...
+    end if
+
+    block
+      integer(c_int) :: new_index_
+      if (present(new_index)) then
+        new_index_ = new_index
+      else
+        new_index_ = 1
+      end if
+
```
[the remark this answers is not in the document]
everythingfunctional: I'm good with adding this as a reminder to ourselves. At some point we *might* be able to address this problem with the smart_pointers library, but I haven't thought it all the way through yet.

src/caffeine/allocation_s.f90
```diff
@@ -28,6 +27,12 @@
...
     me = caf_this_image(current_team%info%gex_team)
+    if (caf_have_child_teams()) then
+      ! Free the child team space to make sure we have space to allocate the coarray
+      if (me == 1) then
+        call caf_deallocate(current_team%info%heap_mspace, current_team%info%child_heap_info%allocated_memory)
+      end if
+    end if
     if (me == 1) then
       handle_size = c_sizeof(unused)
       total_size = handle_size + coarray_size
```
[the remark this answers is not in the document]
everythingfunctional: Yeah, the right thing to do is make proper use of the stat arguments here (or crash if not present), but for now I added a check in `caf_allocate` so at least we won't silently continue.

everythingfunctional:
I found and fixed the bug! Only image 1 was calculating where the new heap started :face_palm:.

src/caffeine/allocation_s.f90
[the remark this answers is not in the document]
bonachea: Thanks for adding the check.

I agree that in order to be PRIF compliant we instead need to return `PRIF_STAT_OUT_OF_MEMORY`, but that can be deferred to another PR. I've entered this as issue #128

test/caf_teams_test.f90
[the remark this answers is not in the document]
bonachea: I agree it's a judgement call. 

In any case this topic is clearly orthogonal the purpose this PR, as those queries are not even implemented yet.

Let's not hold up this PR for this.

bonachea (approved):
LGTM! Thanks for all the work on this!

PR #127, opened 2024-08-07
Put team data in a separate type

everythingfunctional:
- makes it more clear that the components of a team are implementation defined
- makes it easier to default initialize team type components
- also made correct use of initial and current team variables

### DIFF
--- a/src/caffeine/prif_private_s.f90
+++ b/src/caffeine/prif_private_s.f90
@@ -3,8 +3,8 @@
 submodule(prif) prif_private_s
   implicit none
 
-  type(prif_team_type), target :: initial_team
-  type(prif_team_type), pointer :: current_team => null()
+  type(team_data), target :: initial_team
+  type(prif_team_type) :: current_team
   type(c_ptr) :: non_symmetric_heap_mspace
 
   interface
--- a/src/caffeine/prif_private_s.f90
+++ b/src/caffeine/prif_private_s.f90
@@ -3,7 +3,7 @@
 submodule(prif) prif_private_s
   implicit none
 
-  type(prif_team_type), target :: default_team
+  type(prif_team_type), target :: initial_team
   type(prif_team_type), pointer :: current_team => null()
   type(c_ptr) :: non_symmetric_heap_mspace
 
--- a/src/caffeine/program_startup_s.F90
+++ b/src/caffeine/program_startup_s.F90
@@ -11,10 +11,15 @@ contains
     if (prif_init_called_previously) then
        stat = PRIF_STAT_ALREADY_INIT
     else
-       allocate(current_team)
-       call caf_caffeinate(current_team%heap_mspace, current_team%heap_start, non_symmetric_heap_mspace, current_team%gex_team)
-       nullify(current_team%parent_team)
-       nullify(current_team%coarrays)
+       allocate(initial_team%info)
+       call caf_caffeinate( &
+          initial_team%info%heap_mspace, &
+          initial_team%info%heap_start, &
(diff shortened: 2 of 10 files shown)


src/prif.f90
```diff
@@ -1031,4 +1027,12 @@ module subroutine prif_atomic_ref_logical_indirect(image_num, atom_remote_ptr, v
     integer(c_intmax_t) :: lcobounds(15), ucobounds(15)
   end type
 
+  type, private :: team_data
+    type(c_ptr) :: gex_team
+    type(c_ptr) :: heap_mspace
+    integer(c_intptr_t) :: heap_start
+    type(prif_team_type), pointer :: parent_team
```
bonachea, line 1034: I suspect we actually want the `parent_team` component to point directly to the corresponding `team_data` object, otherwise we'll be forced to allocate and leak a second object. Given the `team_data` pointer, a corresponding `prif_team_type` for the parent team can always be manufactured on-demand.

```suggestion
    type(team_data), pointer :: parent_team
```

src/caffeine/prif_private_s.f90
```diff
@@ -3,7 +3,7 @@
 submodule(prif) prif_private_s
   implicit none
 
-  type(prif_team_type), target :: default_team
+  type(prif_team_type), target :: initial_team
   type(prif_team_type), pointer :: current_team => null()
```
bonachea, line 7: Now that `prif_team_type` has been corrected to act as a pointer, we no longer really need the `pointer` and `target` attributes above. 

`current_team` can simply be a (non-pointer) `type(prif_team_type)` with `current_team%info` pointing to the appropriate `team_data`. Similarly for `initial_team`.

```suggestion
  type(prif_team_type) :: initial_team
  type(prif_team_type) :: current_team
```

This will also require a 1-line change in `prif_init` where we initialize `current_team`.

src/caffeine/program_startup_s.F90
```diff
@@ -11,10 +11,15 @@
...
-       nullify(current_team%parent_team)
-       nullify(current_team%coarrays)
+       allocate(initial_team%info)
+       call caf_caffeinate( &
+          initial_team%info%heap_mspace, &
+          initial_team%info%heap_start, &
+          non_symmetric_heap_mspace, &
+          initial_team%info%gex_team)
+       nullify(initial_team%info%parent_team)
+       nullify(initial_team%info%coarrays)
```
bonachea, line 21: We want the above two component initializations (and eventually others) to appear every time Caffeine allocates a `team_data` (which will eventually appear in at least two different places, here and `prif_form_team`).

Wouldn't it make more sense to define initializers for these fields?

bonachea (changes requested):
This PR addresses issue #123 

This looks like the right overall approach. 

I added a few relatively minor suggestions for improvement.

src/caffeine/prif_private_s.f90
bonachea: As a further improvement, we might eventually want to statically allocate the initial team data, ie:

 ```suggestion
  type(team_data), target :: initial_team
  type(prif_team_type) :: current_team 
```

The initial team is "special" for many reasons, especially because so many calls in PRIF now accept an image index that is relative to the initial team (ignoring the current team). This change would allow Caffeine to skip a pointer indirection when doing things like sanity checking that the client provided a valid image index in the initial team.

everythingfunctional:
Thanks for the review @bonachea . All good points, and I've now incorporated all the suggestions.

bonachea (approved):
LGTM!

PR #130, opened 2024-08-28
Add implementation status doc

ktras:


docs/implementation-status.md
```diff
@@ -0,0 +1,245 @@
+# Implementation Status
+
+Caffeine is an implementation of the Parallel Runtime Interface for Fortran (PRIF). This document
+outlines the implementation status in Caffeine of the features defined in the
+[latest PRIF revision](https://dx.doi.org/10.25344/S4WG64). Caffeine contains interfaces for all
```
bonachea, line 5:
```suggestion
[latest PRIF specification, revision 0.4](https://dx.doi.org/10.25344/S4WG64). Caffeine contains interfaces for all
```

docs/implementation-status.md
```diff
@@ -0,0 +1,245 @@
...
+of the PRIF procedures and are callable, but some procedure definitions are unimplemented. For
```
bonachea, line 6:
```suggestion
of the PRIF procedures and the symbols are linkable and callable, but some procedures will fail at runtime with an unimplemented error. For
```

docs/implementation-status.md
```diff
@@ -0,0 +1,245 @@
...
+---
+
+## Image Queries
+### Support = partial
+
+| Procedure | Status | Notes |
+|-----------|--------|-------|
+| `prif_num_images`    | yes | n\a |
+| `prif_num_images_with_team`, `prif_num_images_with_team_number`  | no | n\a |
+| `prif_this_image_no_coarray` | yes | n\a |
```
bonachea, line 58:
```suggestion
| `prif_this_image_no_coarray` | partial | team argument is ignored |
```

docs/implementation-status.md
```diff
@@ -0,0 +1,245 @@
...
+## Collectives
+### Support = partial (...)
+
+| Procedure | Status | Notes |
+|-----------|--------|-------|
+| `prif_co_broadcast` | ... | ... |
+| `prif_co_max`       | ... | ... |
+| `prif_co_min`       | ... | ... |
+| `prif_co_sum`       | ... | ... |
+| `prif_co_reduce`    | ... | ... |
```
bonachea, line 205:
```suggestion
| `prif_co_broadcast` | partial | no support for derived types with `allocatable` components |
| `prif_co_max`       | partial | only supports 32-bit and 64-bit numeric types |
| `prif_co_min`       | partial | only supports 32-bit and 64-bit numeric types |
| `prif_co_sum`       | partial | only supports 32-bit and 64-bit numeric types |
| `prif_co_reduce`    | partial | only supports intrinsic types (no support for derived types) |
```

docs/implementation-status.md
```diff
@@ -0,0 +1,245 @@
...
+support for these optional arguments in Caffeine yet and this is not represented in the tables
+in the following sections.
+
+
+## Program Startup and Shutdown
+### Support = partial
+
+| Procedure | Status | Notes |
+|-----------|--------|-------|
+| `prif_init` | yes | n/a |
```
bonachea, line 45: Looking over the rendered doc, the repeated `n/a` in the Notes column is somewhat visually distracting and makes the tables look "busier" than they are. I suggest we just leave the Notes cell blank when we have nothing to say.

docs/implementation-status.md
```diff
@@ -0,0 +1,245 @@
...
+## Storage Management
+### Support = partial (no support for coarray aliases)
+
+| Procedure | Status | Notes |
+|-----------|--------|-------|
+| `prif_allocate_coarray`    | yes | n/a |
+| `prif_allocate`    | yes | n/a |
+| `prif_deallocate_coarray`    | partial | no `final_func` arg support |
+| `prif_deallocate`    | yes | n/a |
+| `prif_alias_create`    | no | n/a |
```
bonachea, line 75: On a cosmetic note, even in the absence of color we could potentially still make the Status column "pop" a bit more, eg:

| Procedure | Status | Notes |
|-----------|--------|-------|
| `prif_allocate_coarray`    | **YES** |  |
| `prif_allocate`    | **YES** |  |
| `prif_deallocate_coarray`    | *partial* | no `final_func` arg support |
| `prif_deallocate`    | **YES** |  |
| `prif_alias_create`    | no |  |
| `prif_alias_destroy`    | no | |

However if this seems like too much hassle to maintain then we could keep the current convention.

```suggestion
| `prif_deallocate_coarray`    | *partial* | no `final_func` arg support |
| `prif_deallocate`    | **YES** | n/a |
| `prif_alias_create`    | no | n/a |
```

docs/implementation-status.md
```diff
@@ -0,0 +1,245 @@
...
+- [Strided Coarray Access](#Strided-Coarray-Access)
+- [SYNC Statements](#SYNC-Statements)
+- [Locks and Unlocks](#Locks-and-Unlocks)
+- [Critical](#Critical)
+- [Events and Notifications](#Events-and-Notifications)
+- [Teams](#teams)
+- [Collectives](#Collectives)
+- [Atomic Memory Operations](#Atomic-Memory-Operations)
+
+TODO: For features with all no support, flesh out table, but comment it out
```
bonachea, line 26:
```suggestion
```

bonachea (changes requested):
Thanks for the great start on this @ktras !

Reviewing this PR reminds me that many bushels of low-hanging fruit remain available to be "plucked" in Caffeine :-)

I added some updates and suggestions

everythingfunctional:
I agree with Dan's comments, but I don't have any more to add.

docs/implementation-status.md
bonachea: `prif_co_{min,max}` have an implementation for character strings, but @rouson and I have just discovered that it's "fragile" (issue #132) such that we might not want to claim support

bonachea:
@ktras Reminder this is a blocker for our Caffeine release, which we are hoping to finish before the quarterly report due **THIS FRIDAY**.

We should count on at least 1 full day of release finalization, so we really need to iterate this PR towards merge ASAP.

docs/implementation-status.md
ktras: Great point. I have removed the `n/a` notes from the tables.

ktras:
@bonachea Thank you for the feedback and for the reminder of the timeline. I have updated the PR, so please let me know of any further desired changes that may come up in further review.

docs/implementation-status.md
```diff
@@ -198,11 +197,11 @@ in the following sections.
...
 |-----------|--------|-------|
-| `prif_co_broadcast` | ... | ... |
-| `prif_co_max`       | ... | ... |
-| `prif_co_min`       | ... | ... |
-| `prif_co_sum`       | ... | ... |
-| `prif_co_reduce`    | ... | ... |
+| `prif_co_broadcast` | *partial* | no support for derived types with `allocatable` components |
+| `prif_co_max`       | *partial* | only supports 32-bit and 64-bit numeric types |
+| `prif_co_min`       | no        |  |
+| `prif_co_sum`       | no        |  |
```
bonachea, line 203: Why were these two changed from the suggestion? 

I believe all three built-in reduction operations (`co_{min,max,sum}`) are currently in the same state, ie they are believed to be working with gfortran and 32/64-bit numeric datatypes. 

These all have some known problems (e.g. issue #132, which is really a PRIF spec issue).  `co_{min,max}` have some support for Fortran strings but it currently relies on undefined behavior of gfortran. None of these work on flang yet, but we're not publicly claiming flang support for anything yet.

However despite all that I think I'd still characterize all three the same way:

```suggestion
| `prif_co_min`       | *partial* | only supports 32-bit and 64-bit numeric types |
| `prif_co_sum`       | *partial* | only supports 32-bit and 64-bit numeric types |
```

bonachea (changes requested):
Mostly LGTM, just one remaining nit-pick

docs/implementation-status.md
bonachea: Re-reading the prior comment thread I think I understand the miscommunication. 

The concerns with `prif_co_{min,max}` are almost entirely regarding the implementation of character type arguments. IMO 32/64-bit numeric datatypes should be considered working.

ktras: @bonachea Thanks, yes I just misunderstood your earlier comment. Thank you for catching that are for the code suggestion.

bonachea (approved):

PR #131, opened 2024-09-06
Support LLVM Flang (flang-new)

rouson:
This PR 

1. Fixes a code conformance issue that flang identified.
2. Switches a `sync all` statement to `call prif_sync_all()`
3. Updates `install.sh` to add flags required for assumed-rank support if `llvm` is detected in `$FPM_CC --version`.

`fpm test` fails due to a code conformance problem in Veggies.  An issue will be submitted on the Veggies repository.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -397,9 +397,9 @@ exit_if_pkg_config_pc_file_missing "caffeine"
 
 compiler_version=$($FPM_FC --version)
 if [[ $compiler_version == *llvm* ]]; then
-  compiler_flag="-mmlir -allow-assumed-rank -Ofast"
+  compiler_flag="-mmlir -allow-assumed-rank -g -Ofast"
 else
-  compiler_flag="-O3"
+  compiler_flag="-g -O3"
 fi
 
 RUN_FPM_SH="build/run-fpm.sh"
@@ -408,7 +408,7 @@ cat << EOF > $RUN_FPM_SH
 #-- DO NOT EDIT -- created by caffeine/install.sh
 fpm_sub_cmd=\$1; shift
 "${FPM}" "\$fpm_sub_cmd" \\
---profile release \\
+--profile debug \\
 --flag "$compiler_flag" \\
 --compiler "`$PKG_CONFIG caffeine --variable=CAFFEINE_FPM_FC`"   \\
 --c-compiler "`$PKG_CONFIG caffeine --variable=CAFFEINE_FPM_CC`" \\
--- a/install.sh
+++ b/install.sh
@@ -273,7 +273,9 @@ fi
 echo "PKG_CONFIG_PATH=$PKG_CONFIG_PATH"
 
 FPM_FC="$($REALPATH $(command -v $FC))"
-FPM_FC=${FPM_FC/flang-20/flang-new}
+if [[ $FPM_FC == *flang* ]]; then
+  FPM_FC=${FPM_FC/flang-20/flang-new}
+fi
 FPM_CC="$($REALPATH $(command -v $CC))"
 
 ask_package_permission()
--- a/install.sh
+++ b/install.sh
@@ -273,6 +273,7 @@ fi
 echo "PKG_CONFIG_PATH=$PKG_CONFIG_PATH"
 
 FPM_FC="$($REALPATH $(command -v $FC))"
+FPM_FC=${FPM_FC/flang-20/flang-new}
 FPM_CC="$($REALPATH $(command -v $CC))"
 
 ask_package_permission()
--- a/install.sh
+++ b/install.sh
@@ -395,12 +395,21 @@ EOF
 
 exit_if_pkg_config_pc_file_missing "caffeine"
 
+compiler_version=$($FPM_FC --version)
+if [[ $compiler_version == *llvm* ]]; then
+  compiler_flag="-mmlir -allow-assumed-rank -Ofast"
+else
+  compiler_flag="-O3"
+fi
+
 RUN_FPM_SH="build/run-fpm.sh"
 cat << EOF > $RUN_FPM_SH
 #!/bin/sh
 #-- DO NOT EDIT -- created by caffeine/install.sh
 fpm_sub_cmd=\$1; shift
 "${FPM}" "\$fpm_sub_cmd" \\
+--profile release \\
+--flag "$compiler_flag" \\
 --compiler "`$PKG_CONFIG caffeine --variable=CAFFEINE_FPM_FC`"   \\
 --c-compiler "`$PKG_CONFIG caffeine --variable=CAFFEINE_FPM_CC`" \\
 --c-flag "`$PKG_CONFIG caffeine --variable=CAFFEINE_FPM_CFLAGS`" \\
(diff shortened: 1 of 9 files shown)


src/caffeine/collective_subroutines/co_reduce_s.f90
```diff
@@ -269,18 +268,19 @@ subroutine Coll_ReduceSub_c_char(arg1, arg2_and_out, count, cdata) bind(C)
 
       call c_f_pointer(cdata, arglen)
       allocate(character(kind=c_char, len=arglen) :: prototype(count))
```
bonachea, line 270: This variable is now unused and should be removed

```suggestion
```

src/caffeine/collective_subroutines/co_reduce_s.f90
```diff
@@ -260,7 +260,6 @@ subroutine Coll_ReduceSub_c_char(arg1, arg2_and_out, count, cdata) bind(C)
       type(c_ptr), value ::  cdata       !! Client data
 
       character(kind=c_char, len=:), allocatable, target :: prototype(:)
```
bonachea, line 262:
```suggestion
```

bonachea (changes requested):
One minor request, otherwise LGTM

bonachea (approved):
LGTM.

I think we are deliberately omitting documentation for now, at least until we get flang-new into CI and fix the current failures on the co_min/max tests.

test/caf_co_max_test.f90
```diff
@@ -105,12 +105,11 @@ function max_elements_in_3D_string_arrays() result(result_)
...
-      integer i, cyclic_permutation(size(script))
+      integer i, cyclic_permutation(size(script)), me
 
-      associate(me => this_image())
-        associate(cyclic_permutation => [(1 + mod(i-1,size(script)), i=me, me+size(script) )])
-          scramlet = reshape(script(cyclic_permutation), shape(scramlet))
-        end associate
+      call prif_this_image_no_coarray(this_image=me)
+      associate(cyclic_permutation => [(1 + mod(i-1,size(script)), i=me, me+size(script) )])
+        scramlet = reshape(script(cyclic_permutation), shape(scramlet))
```
bonachea, line 112: @rouson CI shows the scramlet test is now failing.

I suspect the problem is the previous incorrect use of `this_image()` was setting `me==1` for all images, but now we actually set it to the image index, and the code fails

bonachea (changes requested):

test/caf_co_max_test.f90
bonachea: One (probably unrelated) bug that jumps out at me here is the code below that does:
```fortran
call prif_co_max(co_max_scramlet, result_image=1)
```
Which leaves `co_max_scramlet` undefined on images other than 1, but the code later inspects the value of `co_max_scramlet` on all images, leading to undefined behavior.

We probably need to drop the `result_image` argument to provide the result to all images.

everythingfunctional:
I *think* the tests that are failing are due to the line `result_ =  assert_that(all(scramlet == co_max_scramlet),"all(scramlet == co_max_scramlet)")`, which should use `expected_scramlet` instead. It may be worth adding 1, 2 and 3d arrays of strings for `assert_equals` to veggies, as showing the actual arrays in the failure message would make this easier to diagnose.

ktras:
> I _think_ the tests that are failing are due to the line `result_ = assert_that(all(scramlet == co_max_scramlet),"all(scramlet == co_max_scramlet)")`, which should use `expected_scramlet` instead. It may be worth adding 1, 2 and 3d arrays of strings for `assert_equals` to veggies, as showing the actual arrays in the failure message would make this easier to diagnose.

This has been fixed on the main branch and I can rebase now or after the changes on this branch have been approved.

bonachea:
@rouson do you want to cherry-pick the commits from PR #142 into this one, or does it make sense to keep them as separate pull requests?

rouson:
> @rouson do you want to cherry-pick the commits from PR #142 into this one, or does it make sense to keep them as separate pull requests?

@bonachea I just tagged you in a comment on PR #142 and merged that one.  @ktras is taking over development on this branch.

@ktras you might want to merge `main` into this branch or rebase this branch off of `main`, which would likely go through cleanly because the new commit that I just merged into `main` is orthogonal to the commits on this branch as far a I recall.

test/caf_co_max_test.f90
rouson: @ktras the comment above by @bonachea likely points us to the problem that the cause of the `prif_co_max` unit test failure.

ktras:
@rouson I think will rebase this branch with `main` once PR #143 is merged in `main`. Because we need those changes on this branch as well.

test/caf_co_max_test.f90
```diff
@@ -148,10 +134,10 @@ function reverse_alphabetize_default_character_scalars() result(result_)
         call prif_co_max(my_word)
       end associate
 
-      call prif_num_images(num_images=num_imgs)
-      associate(expected_word => maxval(words(1:min(num_imgs, size(words)))))
-        result_ = assert_equals(expected_word, my_word)
-      end associate
+      call prif_num_images(num_imgs)
+      !expected_word = maxval(words(1:min(num_imgs, size(words)))) ! this line crashes flang
+      expected_word = "we"
```
bonachea, line 139: Pretty sure this answer is only correct for `num_imgs >= 3` (hence the `min()` in the commented-out version).

test/caf_co_reduce_test.f90
```diff
@@ -39,9 +40,9 @@ function alphabetically_1st_size1_string_array() result(result_)
     end associate
 
     call prif_num_images(num_images=num_imgs)
-    associate(expected_name => minval(names(1:min(num_imgs, size(names)))))
-      result_ = assert_that(all(expected_name == my_name))
-    end associate
+    !expected_name = minval(names(1:min(num_imgs, size(names)))) ! this exposes a flang bug
+    expected_name = "betty"
```
bonachea, line 44: Similarly here, I believe this answer assumes `num_imgs >= 4`.

Can we rewrite the commented-out test code to avoid the flang bug while still preserving the appropriate logic?

bonachea (changes requested):
Noted what looks like some scale-dependent regressions in the test code.

Otherwise LGTM

install.sh
```diff
@@ -395,12 +398,21 @@ EOF
 
 exit_if_pkg_config_pc_file_missing "caffeine"
 
+compiler_version=$($FPM_FC --version)
+if [[ $compiler_version == *llvm* ]]; then
+  compiler_flag="-mmlir -allow-assumed-rank -g -Ofast"
```
ktras, line 403: With the latest versions of flang, this is not necessary. However, maybe we should keep it in for now until I verify that the container of flang isn't a version that we need these flags for. The container was updated 7 days ago, so likely these flags aren't necessary for the CI version of flang.

bonachea (approved):
code LGTM and appears to be passing gfortran CI now

ktras (approved):
LGTM. Awesome and thanks @everythingfunctional

Issue #101, opened 2024-04-01
Expand install.sh to allow passing configure options to GASNet


bonachea:
Currently install.sh configures GASNet-EX with only three hard-coded "customization points":

1. C Compiler
2. C++ compiler
3. Install prefix

ie:
```
FC="$FC" CC="$CC" CXX="$CXX" ./configure --prefix "$PREFIX"
```

I'd consider these the "bare minimum" settings necessary to get a functional GASNet-EX install, and they happen to be sufficient for smp-conduit (no network) on most systems. However this implicitly instructs GASNet's configure to use the auto-detected defaults for all other settings, for which there are several hundred (see GASNet `configure --help`).

Distributed systems differ wildly in the details of available network hardware and especially job spawning, such that many real deployments with a network will require additional configure arguments to override the automatic defaults with system-appropriate settings to get a working/optimal GASNet install.

This issue is to request a mechanism for the command-line invocation of install.sh to pass configure arguments to the underlying GASNet configure. Options include:

1. My recommendation would be the [approach taken by UPC++](https://bitbucket.org/berkeleylab/upcxx/wiki/INSTALL.md#markdown-header-1-configuring-upc) whereby any unrecognized command-line options to `install.sh` are passed-through unmodified to GASNet configure. This mimics the way that configure itself operates with sub-packages.
2. Another (complementary) mechanism would be to define an envvar like `$GASNET_CONFIGURE_ARGS` that supplies a list of configure options for GASNet configure. UPC++'s installer also recognizes this variable for legacy reasons, but it can also be useful in some packaging scenarios where one needs to inject configure options without directly modifying a scripted invocation of `install.sh`.

ktras:
I'm leaning towards option 2, since based on discussion with Dan in a meeting, option 2 is likely what we will want in the long term for integration into LLVM CMake build system.

bonachea:
Resolved in PR #134, deploying both proposed mechanisms

Issue #66, opened 2024-02-14
Add ability to use distributed conduits


ktras:
- [ ] Add a distributed conduit to Github CI for testing purposes (probably udp-conduit)
- [ ] Add argument to install script to pick the conduit (have default be smp conduit)

ktras:
Ability to use distributed conduit, whether in the CI or not as is in the task list above, is the most important aspect of this issue.

PR #134, opened 2024-09-10
issue #101: Expand install.sh to allow passing configure options to GASNet

bonachea:
issue #101: Add pass-thru of install.sh arguments to GASNet configure

Now unrecognized install.sh arguments are assumed to be arguments for
GASNet-EX configure and passed through unchanged.

Users can alternatively pass GASNet configure arguments in shell
variable `$GASNET_CONFIGURE_ARGS`, which experience has shown can be
more convenient in some scripted packaging scenarios.

Fixes issue #101

In addition:

* Ensures that Caffeine-driven builds of GASNet always start with a
  clean GASNet build directory. Previously GASNet's config.cache was surviving
  into and affecting subsequent GASNet configure invocations on the
  same top-level Caffeine build dir.

* Updates the default GASNet configure arguments to select the GASNet
  config settings Caffeine requires, and disable unused features to save
  build time/space.

* Temporarily adds a GASNet configure argument to disable MPI
  interoperability support, which is mostly irrelevant for smp-conduit
  and just leads to confusing link-time errors with recent GASNet.
  We'll need to revisit this for expanding Caffeine's conduit support (issue #66)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -4,7 +4,7 @@ set -e # exit on error
 
 print_usage_info()
 {
-    cat << EOF
+    cat <<'EOF'
 Caffeine Installation Script
 
 USAGE:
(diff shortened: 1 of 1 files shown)


bonachea:
I've added the high-priority label because I'd like this considered for inclusion in the upcoming release

ktras (approved):
LGTM

PR #136, opened 2024-09-12
Replace parallel statements with Caffeine calls

rouson:
This PR
1. Replaces several occurrences of `error stop` and `sync all` with the `prif_error_stop` and `prof_sync_all` calls, respectively.
2. Adds a test for `prif_error_stop` with no stop code.
3. Defines a `unit_test_parameters_m` module with constants for use in different parts of the test suite.


example/support-test/error_stop_with_no_code.f90
```diff
@@ -0,0 +1,16 @@
...
+    ,prif_stop &
+    ,prif_error_stop
+  implicit none
+
+  integer init_exit_code
+  logical(kind=c_bool), parameter :: false = .false._c_bool
+
+  call prif_init(init_exit_code)
+  call prif_error_stop(quiet=false) ! a prif_error_stop unit test passes if this line correctly executes error termination
+  call prif_stop(quiet=false, stop_code_char= "program startup failed") ! a prif_error_stop unit test fails if this line runs
```
ktras, line 14:
```suggestion
  call prif_stop(quiet=false) ! a prif_error_stop unit test fails if this line runs
```

src/caffeine/collective_subroutines/unit_test_parameters_m.mod
ktras, line 1: Please remove this `.mod` file that was accidentally added.

src/caffeine/unit_test_parameters_m.mod
ktras, line 1: Please remove this `.mod` file that was accidentally added.

test/caf_error_stop_test.f90
```diff
@@ -11,25 +12,39 @@ function test_prif_this_image() result(tests)
...
     end function
 
-    function check_integer_stop_code() result(result_)
+    function exit_with_no_stop_code() result(result_)
         type(result_t) :: result_
         integer exit_status
 
         call execute_command_line( &
-          command = "./build/run-fpm.sh run --example error_stop_integer_code > /dev/null 2>&1", &
+          command = "./build/run-fpm.sh run --example error_stop_character_code > /dev/null 2>&1", &
```
ktras, line 26: This shouldn't run `error_stop_character_code` in this command. It should run the example that doesn't have a stop code.

ktras (changes requested):
Looks good but needs a couple simple changes.

test/caf_error_stop_test.f90
```diff
@@ -11,30 +12,44 @@ function test_prif_this_image() result(tests)
...
     end function
 
-    function check_character_stop_code() result(result_)
+    function exit_with_integer_stop_code() result(result_)
         type(result_t) :: result_
         integer exit_status
 
         call execute_command_line( &
-          command = "./build/run-fpm.sh run --example error_stop_character_code > /dev/null 2>&1", &
+          command = "./build/run-fpm.sh run --example error_with_stop_integer_code > /dev/null 2>&1", &
```
ktras, line 39: The name of the example program is still wrong, so it is a problem that the tests are passing, because they should theoretically fail? Because fpm should complain that there is no such example?

example/support-test/README.md
```diff
@@ -1,3 +1,3 @@
 Test Support
 ------------
-The programs in this directory intentionally error-terminate to support the `error stop` tests in `test/caf_error_stop.f90`, which use Fortran's `execute_command_line` to run the programms in this directory and check for the expected non-zero stop codes.  Running the tests in this manner enables the Vegetable tests to continue executing in the presence of error-terminating tests.
+The programs in this directory intentionally terminate to support the `stop` and `error stop` unit tests, which use Fortran's `execute_command_line` to run the programs in this directory and to check for the expected non-zero stop codes.  Running the tests in this manner enables the tests to continue executing the child process launced by `execute_command_line` terminates.
```
bonachea, line 3:
```suggestion
The programs in this directory intentionally terminate to support the `stop` and `error stop` 
unit tests, which use Fortran's `execute_command_line` to run the programs in this directory 
and to check for the expected non-zero stop codes.  Running the tests in this manner enables 
the tests to continue executing after the child process launched by `execute_command_line` 
terminates.
```

example/support-test/error_stop_with_character_code.f90
```diff
@@ -0,0 +1,15 @@
...
+     prif_init &
+    ,prif_stop &
+    ,prif_error_stop
+  implicit none
+
+  integer init_exit_code
+  logical(kind=c_bool), parameter :: false = .false._c_bool
+
+  call prif_init(init_exit_code)
+  call prif_error_stop(quiet=false, stop_code_char="") ! a prif_error_stop unit test passes if this line executes error termination
```
bonachea, line 13: A stronger test would pass a non-empty string here, allowing at least a human to audit that the string appears in the output when run manually

```suggestion
  call prif_error_stop(quiet=false, stop_code_char="USER_PROVIDED_STRING") ! a prif_error_stop unit test passes if this line executes error termination
```

example/support-test/stop_with_character_code.f90
```diff
@@ -0,0 +1,15 @@
...
+     prif_init &
+    ,prif_stop &
+    ,prif_error_stop
+  implicit none
+
+  integer init_exit_code
+  logical(kind=c_bool), parameter :: false = .false._c_bool
+
+  call prif_init(init_exit_code)
+  call prif_stop(quiet=false, stop_code_char="") ! a prif_stop unit test passes if this line executes normal termination
```
bonachea, line 13:
```suggestion
  call prif_stop(quiet=false, stop_code_char="USER_PROVIDED_STRING") ! a prif_stop unit test passes if this line executes normal termination
```

example/hello.f90
bonachea, line 1: Putting this comment here because github won't let me comment on a file that has not been changed.

You missed updating test/main.f90, which currently contains a "bare" `stop` statement and an implicit `stop` statement (falling off the main program), both of which should be replaced with an explicit `call prif_stop()`. 

IOW any instance of `end program` in a PRIF-compliant program must be immediately preceded by either `call prif_stop` or `call prif_error_stop`

bonachea (changes requested):
Noted a few minor but important changes I'd like to see.

Not represented in my review is the fact I consider the entire approach taken in the exit testing of invoking `fpm` from within a unit test and inspecting the exit code to be very fragile. I've just entered this as issue #137

ktras (approved):
LGTM with the understanding that after the upcoming release, we will work to improve the error stop and stop tests documented in issue #137.

test/caf_error_stop_test.f90
bonachea, line 1: As just documented in #137 (and visible in the CI results of this PR), it suspect that `./build/run-fpm.sh` is being invoked in this file and test/caf_stop_test.f90 by _every_ image, which means in an 8-image job, there are 8 redundant invocations of `fpm` for each sub-test, resulting in 64 total processes running each stop test. This is neither scalable nor the intent.

I think the fix may be as simple as converting every `call execute_command_line(...)` to instead be something like this:

```fortran
call prif_this_image_no_coarray(this_image=me)
if (me ==  1) ! one representative image launches the sub-job:
    call execute_command_line(...,exitstat = exit_status,cmdstat = command_status,cmdmsg = command_message)
end if
! and broadcasts the results:
call prif_co_broadcast(exit_status, 1)
call prif_co_broadcast(command_status, 1)
call prif_co_broadcast(command_message, 1)
```

bonachea (commented):
Thanks for merging the changes. IMO overall this PR is definitely an improvement, and I'd like to see it in the release.

I've identified one other serious defect in the exit tests modified by this PR, which are not working as intended even in shared memory. The same defect also exists in main, so this is not technically a regression.

Given the impending release and the fundamental problems documented in #137 that make this entire approach to exit-testing a non-starter in general, fixing this particular defect is basically "putting lipstick on a pig". So I'm also fine ignoring the problem for now and merging as-is.

test/main.f90
```diff
@@ -1,8 +1,17 @@
 ! Generated by cart. DO NOT EDIT
```
bonachea, line 1:
```suggestion
! This file was originally generated by cart, but then manually edited. 
! DO NOT REGENERATE THIS FILE!
```

bonachea (approved):

PR #138, opened 2024-09-12
Release related chores

ktras:


bonachea (commented):
This change is necessary but not sufficient to complete the in-repo changes called for in the [release procedure](https://github.com/BerkeleyLab/caffeine/blob/main/docs/README-release.md).

We also need to update `GASNET_VERSION` in install.sh. Do you plan to do that separately?

ktras:
> This change is necessary but not sufficient to complete the in-repo changes called for in the [release procedure](https://github.com/BerkeleyLab/caffeine/blob/main/docs/README-release.md).
> 
> We also need to update `GASNET_VERSION` in install.sh. Do you plan to do that separately?

Yes, I planned to do it separately, based on the way release procedure is structured and ordered now. However, I agree with you that it can be part of this PR and I have made changes to the release procedure so that after this release, I will update and group the hardcoding of the gasnet version with the other chores related to editing files in the repo.

everythingfunctional (approved):

bonachea (approved):

PR #142, opened 2024-09-16
Define feature-based macros for use via #include

rouson:
This PR defines ceates a new `include/` subdirectory initially containing one `language-support.F90` file defining C-preprocessor macros defined around language feature support.  The two initial macros, `HAVE_SELECTED_LOGICAL_KIND` and `HAVE_PROCEDURE_ACTUAL_FOR_POINTER_DUMMY`.  The `prif.f90` file now contains `#include "language-support.F90"` and uses the first of the two former macros.  The second macro is not currently used, but will be of use in the test suite if it is decided to switch to Julienne in the future.

### DIFF
--- a/include/language-support.F90
+++ b/include/language-support.F90
@@ -1,0 +1,23 @@
+! Copyright (c), The Regents of the University of California
+! Terms of use are as specified in LICENSE.txt
+
+#ifndef HAVE_SELECTED_LOGICAL_KIND
+  ! Define whether the compiler supports standard intrinsic function selected_logical_kind(), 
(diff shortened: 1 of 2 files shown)


rouson:
@bonachea because we wrote this together in screen-share, I'll self-approve it and merge it. I've renamed it from "Build with flang" to the current name because @ktras reminded me that the `fix-stop` branch is further along in supporting `flang-new`. She is taking over development on that branch, which will also include some fixes for issues that she and I discovered with the current implementation of error termination with `character` stop codes.

Issue #78, opened 2024-03-14
Add support for `quiet` argument to `prif_stop` and `prif_error_stop`


ktras:
Add support for `quiet` argument to `prif_stop` and `prif_error_stop`

Issue #77, opened 2024-03-14
Look into Caffeine implementations of `error stop` and `stop`


ktras:
Caffeine implementations of `error stop` and `stop` may need to be updated since it seems to call back and forth between Fortran, then C, then Fortran, which may not be advisable.

bonachea:
The current messy implementation of error_stop also appears to currently be seg-faulting inside `inner_caf_error_stop_character` when compiled by flang-new

PR #143, opened 2024-09-17
Remove pure from `prif_error_stop` and update both `prif_error_stop` and `prif_stop` to support PRIF semantics

ktras:
`prif_error_stop` in the PRIF spec is not `pure`. The caffeine implementation is `pure` since it was `pure` since before PRIF existed. This PR removes the `pure` attribute and makes the jump back and forth across Fortran and C in the implementation of `prif_error_stop` unnecessary, which caused a runtime error when trying to build with LLVM Flang. The jump back and forth across Fortran and C was necessary to obfuscate the non pure behavior (writing to the output_unit) inside of a procedure decorated with the `pure` attribute.

Also, update both `prif_error_stop` and `prif_stop` to support PRIF semantics, which includes implementing the `quiet` arg semantics.

Closes issue #77.
Closes issue #78.

### DIFF
--- a/src/caffeine/program_termination.c
+++ b/src/caffeine/program_termination.c
@@ -1,5 +1,0 @@
-extern void inner_caf_error_stop_integer(void *);
-
-extern void caf_error_stop_integer_c(void *stop_code) {
-    inner_caf_error_stop_integer(stop_code);
-}
--- a/src/caffeine/program_termination.c
+++ b/src/caffeine/program_termination.c
@@ -1,9 +1,4 @@
-extern void inner_caf_error_stop_character(void *, int);
 extern void inner_caf_error_stop_integer(void *);
-
-extern void caf_error_stop_character_c(void *stop_code, int length) {
-    inner_caf_error_stop_character(stop_code, length);
-}
 
 extern void caf_error_stop_integer_c(void *stop_code) {
     inner_caf_error_stop_integer(stop_code);
--- a/src/caffeine/program_termination_s.f90
+++ b/src/caffeine/program_termination_s.f90
@@ -110,28 +110,6 @@ contains
 
   end subroutine
 
-  function f_c_string(f_string) result(c_string)
-    character(len=*), intent(in) :: f_string
-    character(len=1,kind=c_char) :: c_string(len(f_string))
-
-    integer :: i
-
-    do concurrent (i = 1:len(f_string))
-      c_string(i) = f_string(i:i)
-    end do
-  end function
-
-  function c_f_string(c_string, length) result(f_string)
-    integer(c_int), intent(in) :: length
-    character(len=1,kind=c_char), intent(in) :: c_string(length)
-    character(len=length) :: f_string
-
-    integer :: i
-    do concurrent (i = 1:length)
-      f_string(i:i) = c_string(i)
-    end do
-  end function
-
   module procedure prif_fail_image
     call unimplemented("prif_fail_image")
   end procedure
--- a/src/caffeine/program_termination_s.f90
+++ b/src/caffeine/program_termination_s.f90
@@ -26,8 +26,7 @@ contains
 
       call prif_sync_all
 
-      !write(output_unit, *) "caf_stop: stop code '", stop_code, "'"
-      write(output_unit, *) stop_code
+      write(output_unit, *) "STOP ", stop_code
       flush output_unit
 
       if (.not. present(stop_code)) call caf_decaffeinate(exit_code=0_c_int) ! does not return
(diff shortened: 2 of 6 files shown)


rouson:
@bonachea see the @ktras comment above.  We're no longer engaged in trickery to circumvent standard-conformance.

src/caffeine/program_termination_s.f90
```diff
@@ -125,31 +89,8 @@ subroutine inner_caf_error_stop_integer(stop_code) bind(C, name = "inner_caf_err
...
-  pure function c_f_string(c_string, length) result(f_string)
-    integer(c_int), intent(in) :: length
-    character(len=1,kind=c_char), intent(in) :: c_string(length)
-    character(len=length) :: f_string
-
-    integer :: i
-    do concurrent (i = 1:length)
-      f_string(i:i) = c_string(i)
-    end do
-  end function
```
bonachea, line 151: We may end up needing these string-conversion utilities later (e.g. to fix the implementation of co_min/co_max on strings), but we can always restore them when needed (and they really belong in prif_private_s.f90 instead of here).

src/caffeine/program_termination_s.f90
```diff
@@ -62,61 +62,25 @@ end subroutine prif_stop_character
...
-
     integer exit_code
 
     if (.not. present(stop_code)) then
-
       call caf_decaffeinate(exit_code=1)
-
     else if (stop_code==0) then
-
       write(error_unit) stop_code
```
bonachea, line 84: IIUC there's an inconsistency here: the integer error stop code is being written to error_unit only if and only if it's zero. This not part of the spec, and I suspect it arose from the current implementation approach where `prif_error_stop_character` calls `prif_error_stop_integer`.

Of course the actual spec says the `quiet` argument (which is currently ignored by the implementation) controls output behavior, so in that sense both current behaviors are "wrong".

We could consider just fixing #78 along with this PR since it should only require a few extra lines of code in these same subroutines, but I'm also fine with deferring that.

src/caffeine/program_termination.c
bonachea, line 1: This PR addresses #77, and that information really should appear in at least one commit message.

As an extra bonus, if you include a string like `fixes #77` in a commit message, then github will automatically close the corresponding issue when that commit is merged to main.

I recommend we all get in the habit of using this [github feature](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue), because it helps manage our issue tracker and also leaves a permanent record in git history to any issue tracker discussions relevant to a given commit.

bonachea (changes requested):
I'm strongly in favor of this improvement. Thanks @ktras !

I made one cosmetic request that can be resolved by amending a commit message.

I also noted that the (error) stop implementation still diverges from the PRIF spec (#78), but I'm fine also deferring the remaining work to a future PR.

src/caffeine/program_termination_s.f90
ktras: Addressed in this PR. Please check the implementation when you are able. I believe this PR should now match the PRIF semantics, including the semantics of the `quiet` arg.

src/caffeine/program_termination.c
ktras: Commit message was amended to use `fixes`.

src/caffeine/program_termination_s.f90
```diff
@@ -50,40 +62,44 @@ end subroutine prif_stop_character
...
-  subroutine prif_error_stop_character(stop_code)
+  subroutine prif_error_stop_character(quiet, stop_code)
     !! stop all images and provide the stop_code as the process exit status
+    logical(c_bool), intent(in) :: quiet
     character(len=*), intent(in) :: stop_code
 
-    write(error_unit, *) stop_code
-    flush error_unit
+    if (.not. quiet) then
+      write(error_unit, *) stop_code
```
bonachea, line 78: IIUC the details of what output formatting are unspecified by Fortran (and PRIF), but for symmetry with `prif_stop_character` and `prif_error_stop_integer` shouldn't this be something like:
```suggestion
      write(error_unit, *) "ERROR STOP '" // stop_code // "'"
```

bonachea (changes requested):
Thanks for the further improvements @ktras !
This is a huge improvement over the stop routines currently in main.

I noted one more minor nitpick, otherwise this is probably ready to merge.

On a related topic, on a non-quiet `stop`/`prif_stop`, Caffeine currently prints output from *every* image, which could become tedious at large scale. I suggest we enter an enhancement issue to do something "smarter" here. E.g. F2023 11.4 suggests:

> NOTE 1 When normal termination occurs on more than one image, it is expected that a processor-dependent summary of any stop codes and signaling exceptions will be made available

src/caffeine/program_termination_s.f90
ktras: Thanks for the catch, I did want that change and symmetry, but I forgot it. Appreciate it!

PR #146, opened 2024-12-10
Update `prif_allocate_coarray` to match PRIF v0.5

ktras:


src/caffeine/allocation_s.f90
```diff
@@ -20,12 +20,10 @@
     integer :: me
     type(c_ptr) :: whole_block
     integer(c_ptrdiff_t) :: block_offset
-    integer(c_size_t) :: handle_size, coarray_size, total_size
+    integer(c_size_t) :: handle_size, total_size
```
bonachea, line 23: Weak request to rename the misleading variable `handle_size` to `descriptor_size`, because that's what it actually represents.

src/caffeine/allocation_s.f90
```diff
@@ -20,12 +20,10 @@
...
     type(handle_data) :: unused
     type(handle_data), pointer :: unused2(:)
```
bonachea, line 25: On a related note, PRIF 0.5 mandates that we rename `handle_data` to `prif_coarray_descriptor`, so it might make sense to make that related change in this same PR (especially given both changes are directly relevant to our expected near-term consumer).

bonachea (changes requested):
Looks like a great start! 
Made a few relatively minor suggestions.

bonachea (approved):
Changes LGTM.

Please rebase to pick up the changes from PR #148, fix the semantic conflict and ensure CI is passing before merge.

everythingfunctional (approved):

bonachea (approved):

PR #147, opened 2024-12-10
Update MacOS version in CI

ktras:


everythingfunctional (approved):
I made a couple of attempts to go forward more than one version, but this works for now I guess.

PR #148, opened 2024-12-10
implement prif_local_data_pointer

everythingfunctional:


bonachea (changes requested):
Please add at least one use of the new entry point somewhere in test code.

Otherwise LGTM.

bonachea (approved):
LGTM! 

Note this PR has a semantic (likely non-textual) conflict with PR #146 and whichever is merged second will need to be adjusted before merge.

PR #149, opened 2024-12-17
Replace `c_intmax_t` with `c_int64_t`

### LINKED ISSUE #150: Replace `c_intmax_t` with `c_int64_t`
According with changes in the PRIF interfaces in PRIF v0.5, replace all kinds `c_intmax_t` with `c_int64_t` and on the C side, also replace `intmax_t` with `int64_t`.

ktras:
Closes #150.

### DIFF
--- a/src/caffeine/caffeine.c
+++ b/src/caffeine/caffeine.c
@@ -287,7 +287,7 @@ size_t caf_elem_len(CFI_cdesc_t* a_desc)
   return a_desc->elem_len;
 }
 
-void caf_form_team(gex_TM_t current_team, gex_TM_t* new_team, intmax_t team_number, int new_index)
+void caf_form_team(gex_TM_t current_team, gex_TM_t* new_team, int64_t team_number, int new_index)
 {
   gex_TM_Split(new_team, current_team, team_number, new_index, NULL, 0, GEX_FLAG_TM_NO_SCRATCH);
 }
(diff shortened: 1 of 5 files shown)


everythingfunctional (approved):

src/caffeine/caffeine.c
```diff
@@ -287,7 +287,7 @@ size_t caf_elem_len(CFI_cdesc_t* a_desc)
   return a_desc->elem_len;
 }
 
-void caf_form_team(gex_TM_t current_team, gex_TM_t* new_team, intmax_t team_number, int new_index)
+void caf_form_team(gex_TM_t current_team, gex_TM_t* new_team, int64_t team_number, int new_index)
 {
   gex_TM_Split(new_team, current_team, team_number, new_index, NULL, 0, GEX_FLAG_TM_NO_SCRATCH);
```
bonachea, line 292: The third argument to `gex_TM_Split` has type `int`, so this call involves an implicit narrowing conversion of `team_number` from 64 to 32 bit, which will silently truncate anything in the upper bits.

Caffeine will not be handling this in the near future, so I'd like an assertion here, something of the form:

```suggestion
   // GASNet color argument is int (32-bit), check for value truncation:
  assert((unsigned int)team_number == team_number);
  gex_TM_Split(new_team, current_team, team_number, new_index, NULL, 0, GEX_FLAG_TM_NO_SCRATCH);
```

bonachea: This is not directly relevant to this PR, just an undocumented issue that's been on my radar for awhile.

bonachea (changes requested):
Made one minor tangential request, otherwise LGTM

bonachea (approved):

PR #158, opened 2024-12-19
Update collective subroutines to match PRIF 0.5

### LINKED ISSUE #69: Rework `prif_co_reduce` implementation
Rework the `prif_co_reduce` implementation to stop using the current strategy involving the abstract interfaces for the various C interoperable types. This may be a prerequisite to supporting non-blocking `co_reduce`.

### LINKED ISSUE #132: Consider eliminating `type(*)` in the collective subroutines
Fortran 2023 18.3.7 item 7 in the list on page 515 prohibits assumed-type, assumed-rank arguments without the `value` attribute.

### LINKED ISSUE #144: Fix definitions for `float_Complex_workaround`, `double_Complex_workaround` in `caffeine.c`
The definitions are based on a macro and the logic with the macro needs to be fixed.

### LINKED ISSUE #151: Update `prif_co_min` to match PRIF v0.5
- [ ] `prif_co_min`
- [ ] `prif_co_min_character`

### LINKED ISSUE #152: Update `prif_co_max` to match PRIF v0.5
- [ ] `prif_co_max`
- [ ] `prif_co_max_character`

### LINKED ISSUE #153: Update `prif_co_sum` to match PRIF v0.5

### LINKED ISSUE #154: Update `prif_co_reduce` to match PRIF v0.5

### LINKED ISSUE #155: Update `prif_co_broadcast` to match PRIF v0.5

### LINKED ISSUE #159: Implement prif_co_{min,max,sum} for all interoperable ints and reals
PRIF 0.5 says `prif_co_{min,max}` argument: 
> shall be of type integer with an interoperable kind or of type real with an interoperable kind

and `prif_co_sum` argument:
> shall have any numeric type with an interoperable kind

However based on source inspection of caffeine.c:
* `caf_co_{min,max}` only supports: int32, int64, float, double
* `caf_co_sum` only supports: int32, int64, float, double, float complex, double complex

We should add branches to invoke `prif_co_reduce` for the remaining cases:
* `caf_co_{min,max}` missing: int8, int16, long double
* `caf_co_sum` missing: int8, int16, long double, long double complex

We should also add logic to ensure that aliased kind parameters like `CFI_type_size_t`, `CFI_type_long_long`, `CFI_type_int_least64_t`, `CFI_type_short`, `CFI_type_signed_char` etc all resolve to the appropriate logic for the four fixed-width integer basic types.

everythingfunctional:
### Description

Updates the collective interfaces and implementations to match the latest PRIF 0.5 revision.

Largely rewrites all the relevant test code.

This PR was begun by @everythingfunctional and has been continued by @bonachea.

Fixes #69
Fixes #132
Fixes #144
Fixes #151
Fixes #152
Fixes #153 
Fixes #154 
Fixes #155
Fixes #159

### Status:
* Ready for review!
* Passes testing on Flang {19.1.1, 20.1.0}, gfortran {13.3, 14.2}
* The (deprecated) Flang `-Ofast` option has been replaced with `-O3`, and flags specific to flang-19 have been hoisted out of install.sh into CI input.
* One co-reduce subtest remains disabled, due to missing compiler support for parameterized derived types in both flang 20 and gfortran 14.2. This is mostly notable because this test is (a simplified version of) the example code presented in the `prif_co_reduce` section of PRIF 0.5 spec. However there's probably nothing further we can do about this other than filing compiler bugs.
* `prif_co_{min,max,sum}` do not support `long double` (or `long double complex`), nor do they support integer types wider than 64-bit. However those types seem sufficiently unimportant in practice that I think we should just open an issue and defer it indefinitely.


rouson:
@everythingfunctional flang supports kind type parameters for derived types but not length type parameters. Does Caffeine use length type parameters?

everythingfunctional:
> Does Caffeine use length type parameters?

One of the tests was trying to demonstrate how `co_reduce` could be used with derived types with length type parameters. Otherwise Caffeine does not use length type parameters.

bonachea (commented):
Looks good so far, although there's obviously still a ways to go before it reaches feature parity with PRIF 0.4 / main

bonachea:
Brad's version at 171bf087be6f4758c9616b9b5560736267e96424 rebased onto main and conflicts resolved manually

src/caffeine/collective_subroutines/co_max_s.F90
```diff
@@ -1,36 +1,65 @@
...
+  module procedure prif_co_max_character
+    integer(c_size_t), target :: char_len
+    procedure(prif_operation_wrapper_interface), pointer :: op
 
+    char_len = len(a)
+    op => char_max_wrapper
+#if defined(__GFORTRAN__) && 0
+    ! gfortran 13.2 (sometimes?) crashes on the call below
+    ! internal compiler error: in make_decl_rtl, at varasm.cc:1442
+    call unimplemented("prif_co_max_character")
```
bonachea, line 59: I never did figure out the exact conditions that cause this ICE, which is why I left this #if branch here but statically disabled.

In subsequent testing gfortran-13 seems to be stable using the implementation below, so I suspect there was a second contributing factor to the ICE elsewhere in the code that I never discovered and has since changed.

bonachea:
Now also fixes #159, completing the Collectives Subroutines

bonachea:
Rebased to pickup CI changes on `main`

install.sh
```diff
@@ -417,7 +417,7 @@ user_compiler_flags="${CPPFLAGS:-} ${FFLAGS:-}"
 
 compiler_version=$($FPM_FC --version)
 if [[ $compiler_version == *llvm* ]]; then
-  compiler_flag="-mmlir -allow-assumed-rank -g -Ofast"
+  compiler_flag="-g"
```
bonachea, line 420: The flang optimizer bug being worked-around here is now documented in issue #205

bonachea: And workaround added in d17b2cde

src/caffeine/co_broadcast_s.F90
```diff
@@ -0,0 +1,30 @@
...
+  subroutine contiguous_co_broadcast(a, source_image, stat, errmsg, errmsg_alloc)
+    type(*), intent(inout), target, contiguous :: a(..)
+    integer(c_int), intent(in) :: source_image
+    integer(c_int), intent(out), optional :: stat
+    character(len=*), intent(inout), optional :: errmsg
+    character(len=:), intent(inout), allocatable, optional :: errmsg_alloc
+
+    if (present(stat)) stat=0
+    call caf_co_broadcast(a, source_image, product(shape(a)), current_team%info%gex_team)
+      ! With a compliant Fortran 2018 compiler, pass in c_sizeof(a) as the `Nelem` argument
```
ktras, line 26: Maybe outdated comment here and on the next line.

src/caffeine/caffeine.c
```diff
@@ -28,12 +29,12 @@ static gex_TM_t myworldteam;
 typedef void(*final_func_ptr)(void*, size_t) ;
 typedef uint8_t byte;
 
-#if __GNUC__ >= 12
-  #define float_Complex_workaround  CFI_type_float_Complex
-  #define double_Complex_workaround CFI_type_double_Complex
-#else
+#if PLATFORM_COMPILER_GNU && PLATFORM_COMPILER_VERSION_LT(12,0,0)
```
ktras, line 32: Since we don't claim support for gfortran 11, how about we remove this workaround macros?

src/caffeine/caffeine.c
```diff
@@ -204,9 +205,18 @@ void caf_sync_team( gex_TM_t team ) {
...
-)
-{
+  CFI_cdesc_t* a_desc, int result_image, size_t num_elements, gex_Coll_ReduceFn_t user_op, void* client_data, gex_TM_t team
+) {
+  assert(a_desc);
+  assert(result_image >= 0);
+  assert(num_elements > 0);
+  assert(user_op);
+#if PLATFORM_COMPILER_GNU
+  // gfortran 13.2 c_funloc is non-compliant
```
ktras, line 215:
```suggestion
  // gfortran 13.2 & 14 - c_funloc is non-compliant
```

test/main.F90
ktras, line 1: Yay!

src/caffeine/co_broadcast_s.F90
bonachea: I was incorrect in our meeting today because I hadn't looked carefully at `caf_co_broadcast` lately. 

Now that I have, I believe this comment remains accurate, although I'm dubious whether the suggested optimization would result in any measurable difference in performance.

src/caffeine/caffeine.c
bonachea: Done in 3f12d793

src/caffeine/co_broadcast_s.F90
ktras: Okay, thanks for the update. If its not going to help, maybe the suggestion won't ever be adopted, even with a Fortran 2018 compliant compiler, but I don't mind keeping the comment for now.

ktras (approved):
LGTM. Thanks for the response to feedback.

src/caffeine/collective_subroutines/co_max_s.F90
ktras: Interesting and unfortunate, but good choice to have this statically disabled but there in case we run into the bug again, this should hopefully save time in figuring out what is going on.

Issue #163, opened 2024-12-23
caf_co_min and caf_co_max tests fail when num_images()==1


rouson:
Building with GCC 14.2.0 on macOS  14.6.1 and executing the test suite in a single image reports failures in `caf_co_min` and `caf_co_max`:
```
git checkout 0.4.0-rc1
./install.sh
export GASNET_PSHM_NODES=1
./build/run-fpm.sh test
```
which produces the trailing following output:
```
Test that
    The prif_co_max subroutine computes the maximum
        elements across images with 2D arrays of strings
            Expected to be true
                User Message:
                    |all(expected == scramlet)|
    The prif_co_min subroutine computes the minimum
        elements across images with 2D arrays of strings
            Expected to be true
                User Message:
                    |all(expected == scramlet)|

2 of 59 cases failed
2 of 84 assertions failed
```
The test suite passes when `GASNET_PSHM_NODES` is 2 or 7.

rouson:
I think the test logic correct. Using `git bisect` leads to 34687f25f500870a9db035edc3d62c93a252d53a as the  first bad commit.  However, the number of test failures varies between 2 and 3 along the way, which might indicate more complex behaviors than the assumptions implied by `git bisect`.

PR #168, opened 2024-12-28
Fix issue #163

rouson:
This PR 
* Fixes an issue that caused test failures in single-image runs of the `prif_co_min` and `prif_co_max` tests,
* Adds clarifying comments in the `prif_co_min` and `prif_co_max` tests, and
* Adds a missing format in the `write` statement in `prif_stop_integer`.

### DIFF
--- a/src/caffeine/program_termination_s.f90
+++ b/src/caffeine/program_termination_s.f90
@@ -106,13 +106,13 @@ contains
 
     if (present(stop_code)) then
       if (.not.quiet) then
-        write(error_unit) "ERROR STOP ", stop_code
+        write(error_unit,'(a)') "ERROR STOP ", stop_code
         flush error_unit
       end if
       exit_code = stop_code
     else
       if (.not.quiet) then
-        write(error_unit) "ERROR STOP"
+        write(error_unit,'(a)') "ERROR STOP"
         flush error_unit
       end if
       exit_code = 1_c_int
--- a/src/caffeine/program_termination_s.f90
+++ b/src/caffeine/program_termination_s.f90
@@ -106,7 +106,7 @@ contains
 
     if (present(stop_code)) then
       if (.not.quiet) then
-        write(error_unit,'(a)') "ERROR STOP ", stop_code
+        write(error_unit,'(A, I0)') "ERROR STOP ", stop_code
         flush error_unit
       end if
       exit_code = stop_code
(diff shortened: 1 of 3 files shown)


test/caf_co_max_test.f90
```diff
@@ -104,21 +104,24 @@ function max_double_precision_2D_array() result(result_)
...
+      character(len=len(script)), dimension(size(script,1),size(script,2)) :: slice
       integer me, ni
 
       call prif_this_image_no_coarray(this_image=me)
       call prif_num_images(ni)
-      scramlet = script(:,:,mod(me,size(script,3))+1)
-      call prif_co_max(scramlet)
-      expected = maxval(script(:,:,1:min(ni,size(script,3))), dim=3)
-      result_ = assert_that(all(expected == scramlet),"all(expected == scramlet)")
+      associate(slice_number => mod(me-1,size(script,3)) + 1)
```
bonachea, line 118: It seems like the `me-1` change here was the crucial fix for the defect. Is that correct?

src/caffeine/program_termination_s.f90
```diff
@@ -106,13 +106,13 @@ subroutine prif_error_stop_integer(quiet, stop_code)
 
     if (present(stop_code)) then
       if (.not.quiet) then
-        write(error_unit) "ERROR STOP ", stop_code
+        write(error_unit,'(a)') "ERROR STOP ", stop_code
```
bonachea, line 109: Shouldn't this format statement also include an entry for `stop_code`?

```suggestion
        write(error_unit,'(A, I5)') "ERROR STOP ", stop_code
```

I'm probably missing something here.

bonachea (commented):
This PR fixes test code that targets PRIF 0.4, and will be incompatible with PRIF 0.5 where the interfaces have been renamed to `prif_co_{min,max}_character` and require `kind=c_char`.

PR #158 contains the work-in-progress to convert the collectives implementation and this same test code to PRIF 0.5, and this PR will conflict with that work.

That being said, I have no objection to merging this as a near-term fix until the PRIF 0.5 collectives work is ready.

test/caf_co_max_test.f90
rouson: @bonachea correct.

src/caffeine/program_termination_s.f90
rouson: @bonachea good catch.  I'm switching `I5` to `I0`, which lets the compiler select the `integer` width. 
```suggestion
        write(error_unit,'(A, I0)') "ERROR STOP ", stop_code
```

bonachea (approved):

Issue #137, opened 2024-09-12
Unit tests for `prif_stop` and `prif_error_stop` make fragile non-portable assumptions


bonachea:
Currently the approach taken to unit testing `prif_stop` and `prif_error_stop` is to unconditionally invoke `./build/run-fpm.sh` in the `fpm` built Caffeine unit test, and inspecting the resulting process exit code. 

I consider this entire approach to be very fragile for multiple reasons:

1. Assumes Caffiene test executable is run from the source/build directory
2. Assumes `fpm` (and possibly the compiler) are available on the compute node
3. Assumes `fpm` is capable of launching parallel jobs at all
4. Assumes parallel jobs can be launched at all (by any command) from the compute node
5. Currently appears to have EVERY image launch the subjob
6. Relies on process exit code propagation, which can be unreliable in loosely coupled distributed systems
 
I expect one or more of the above assumptions to be violated on some systems (completely breaking the Caffeine unit test) once we incorporate distributed conduits and non-trivial job spawners.

As such that we'll eventually need a "kill switch" to disable this practice, or better yet a more robust approach to exit testing that doesn't rely on programmatically invoking fom to spawn a sub-job.

bonachea:
Direct evidence that subjob invocations of `fpm` are not being invoked the correctly once-per-test, but rather once-per-image-per-test (problem 5 listed above):

```
{pcp-d-10} env CC=gcc CXX=c++ FC=gfortran GASNET_PSHM_NODES=1 ./build/run-fpm.sh test | tail -n 50
Project is up to date
        sums integer(c_int64_t) scalars with no optional arguments present
        multiplies default real scalars with all optional arguments present
        multiplies real(c_double) scalars with all optional arguments present
        performs a collective .and. operation across logical scalars
        sums default complex scalars with a stat-variable present
        sums complex(c_double) scalars with a stat-variable present
        sums default integer elements of a 2D array across images
    The prif_co_sum subroutine
        sums default integer scalars with no optional arguments present
        sums default integer scalars with all arguments present
        sums integer(c_int64_t) scalars with stat argument present
        sums default integer 1D arrays with no optional arguments present
        sums default integer 15D arrays with stat argument present
        sums default real scalars with result_image argument present
        sums double precision 2D arrays with no optional arguments present
        sums default complex scalars with stat argument present
        sums double precision 1D complex arrays with no optional arguments present
    A program that executes the prif_error_stop function
[...]

PR #133, opened 2024-09-08
Unit test with Julienne

rouson:
Because the Veggies unit testing framework exposes what appears to be an internal compiler error in LLVM Flang (`flang-new`), this PR refactors the Caffeine test suite to replace Veggies with [Julienne], which provides similar functionality and supports Flang. This commit also removes the unused iso_varying_string dependency and renames all `test/caf_*` files to `test/prif_*`.

To Do
------

- [x] Update `fpm.toml.template`: replace veggies with julienne and remove the unused iso_varying_string dependency
- [x] Replace main test program
- [x] Refactor/rename prif_init test module
- [x] Refactor/rename prif_allocate test module
- [x] Refactor/rename prif_co_broadcast test module
- [x] Refactor/rename prif_co_max test module
- [x] Refactor/rename prif_co_min test module
- [x] Refactor/rename prif_co_reduce test module
- [x] Refactor/rename prif_co_sum test module
- [x] Refactor/rename prif_error_stop test module
- [x] Refactor/rename prif_image_index test module
- [x] Refactor/rename prif_num_images test module
- [x] Refactor/rename prif_rma test module
- [x] Refactor/rename prif_stop test module
- [x] Refactor/rename prif_teams test module
- [x] Refactor/rename prif_this_image test module

[Julienne]: https://github.com/berkeleylab/julienne

PR #169, opened 2024-12-28
Test with Julienne

rouson:
This PR replaces PR #133 and accomplishes the following:
1. Switch from testing with Veggies to testing with Julienne.
2. Make progress on issue #137 by moving the termination tests to a separate `fpm` project in the new `test-termination/` subdirectory as suggested in comment [r1761996357].
3.  Define a new `prif_test_t` type that extends Julienne's `test_t` type and override `test_t`'s `report` type-bound procedure, replacing parallel Fortran features with PRIF calls.

The primary motivation for the switch to Julienne is complexity reduction: the Julienne software stack contains 3850 lines of code versus 30513 in Veggies as determined by executing following command on the current main branch of both repositories:
```
wc -l `find src example test build/dependencies -iname '*.f90'
```
As a consequence, the test suite runs much faster even after accounting for some possible reduction in test coverage because this test suite was originally developed in 3-4 months ago so any tests added to the Veggies-based test suite since then are not represented here.  Determining whether and to what extent the test coverage is actually diminished is on the To Do list below.

To Do
------
- [ ] Edit the CI scripts to run the termination tests in the new `test-termination` subdirectory. 
- [ ] Review and edit the `test/` and `test-termination/` subdirectories to match the test coverage of the previous Veggies-based tests.

[r1761996357]:  https://github.com/BerkeleyLab/caffeine/pull/133#discussion_r1761996357


everythingfunctional (commented):
I haven't reviewed the changes the tests yet, but I have a few higher level comments so far.

1. You need to add an `fpm.toml` file in the `test-termination` folder. At present the instructions in the `test-termination/README.md` do not work. The `fpm.toml` I created is listed below.
2. It is unclear from the output whether the tests pass or fail. I would recommend putting together some sort of script that checks the output of each individual test rather than have users try to inspect the outputs and determine pass/fail manually. This should make getting the CI to work easier as well.
3. I would have made the switch to Julienne a separate PR, as the stop tests don't really depend on the test framework and vice-versa.
4. I think on balance the switch to Julienne might be the right move, but I think it's worth pointing out some of the important differences:
   1. You get less detailed output from running the test suite. i.e. it doesn't report how many or what values are being compared. This could make debugging failing tests (or erroneously passing tests) more difficult.
   2. The framework is simpler, and it is one you control.
5. I think you could adapt `veggies` in a similar way as you did with `julienne`, so if you're curious to try it, I could point you at the few things you'd want/need to overload.

example `fpm.toml` file
```toml
name = "test-termination"
version = "0.4.1"
license = "BSD"
author = ["Damian Rouson", "Brad Richardson", "Katherine Rasmussen", "Dan Bonachea"]
maintainer = "rouson@lbl.gov"
copyright = "2021-2024 UC Regents"

[dependencies]
caffeine = {path = ".."}

[build]
link = ["gasnet-smp-seq", "gcc", "m"]
```

rouson:
@everythingfunctional thanks for the quick review!
 
> 1. You need to add an `fpm.toml` file in the `test-termination` folder. At present the instructions in the `test-termination/README.md` do not work. The `fpm.toml` I created is listed below.

Good catch.  I just pushed my `fpm.toml`, which I mistakenly didn't add before. This required `git add -f` because `fpm.toml` is in the `.gitignore` due to the top-level manifest being generated by `install.sh`.

> 2. It is unclear from the output whether the tests pass or fail. I would recommend putting together some sort of script that checks the output of each individual test rather than have users try to inspect the outputs and determine pass/fail manually. This should make getting the CI to work easier as well.

That's the plan.  I'm hoping @ktras can work on this.

> 3. I would have made the switch to Julienne a separate PR, as the stop tests don't really depend on the test framework and vice-versa.

I agree. If we decide against the switch to Julienne, I'll create a separate PR to replace the old program termination tests with the new ones.

> 4. I think on balance the switch to Julienne might be the right move, but I think it's worth pointing out some of the important differences:
>    
>    1. You get less detailed output from running the test suite. i.e. it doesn't report how many or what values are being compared. This could make debugging failing tests (or erroneously passing tests) more difficult.

This morning's release of Julienne 1.6.0 adds this information in diagnostic output.  This issue has been bugging me for a while, but it took some focused time over the break to think of how to add such diagnostic output in backwards-compatible way.   Of course, there aren't many projects using Julienne yet (two-ish) but it would be nice to not have to rewrite multiple projects' test suites all at once.  Fortunately I finally figured out a backwards-compatible path and that's in Julienne 1.6.0.

>    2. The framework is simpler, and it is one you control.

> 5. I think you could adapt `veggies` in a similar way as you did with `julienne`, so if you're curious to try it, I could point you at the few things you'd want/need to overload.

Very cool. It would probably be useful info for learning purposes either way so pointers are welcome.

bonachea (changes requested):
Made some initial observations based on a quick skim of the high-level changes.

I have not yet dived into review of the test code.

test-termination/fpm.toml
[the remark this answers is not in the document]
rouson: @bonachea if we agree that the new approach is a step in the right direction, then it might make sense to merge the termination-tests as is before more thoroughly addressing #137.  How does that sound?

rouson:
As of commit fcf2fa8, this PR now adds the capability for diagnostic output in the event of a test failure for all tests except those in `test/prif_allocate_test_m.F90 `.  Adding diagnostic output to the latter test will likely require a relatively minor update to Julienne.  That means 48 of 52 tests can produce diagnostic output.  I think that's a sufficiently large fraction that the PR could be merged before adding the diagnostic output in the four missing cases.

manifest/fpm.toml.template
```diff
@@ -6,7 +6,7 @@ maintainer = "rouson@lbl.gov"
 copyright = "2021-2024 UC Regents"
 
 [dev-dependencies]
-veggies = {git = "https://gitlab.com/everythingfunctional/veggies", tag = "v1.1.3"}
-iso_varying_string = {git = "https://gitlab.com/everythingfunctional/iso_varying_string.git", tag = "v3.0.4"}
+julienne = {git = "https://github.com/berkeleylab/julienne.git", tag = "1.5.3"}
+assert = {git = "https://github.com/berkeleylab/assert.git", tag = "2.0.0"}
```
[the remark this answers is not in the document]
bonachea: As agreed in our meeting, I'll be working on this upgrade as a separate PR

test-termination/fpm.toml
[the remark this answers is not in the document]
bonachea: I agree we shouldn't hold up this PR waiting for a complete solution to #137. I've changed the PR description to stop claiming a complete fix to that issue, so that GitHub hopefully won't automatically close it.

I'm not yet prepared to approve a merge of this PR as-is, let's plan to discuss on Wednesday.

test/prif_test_s.f90
```diff
@@ -0,0 +1,76 @@
...
+  use julienne_m, only : command_line_t
+  use prif, only : prif_this_image_no_coarray, prif_co_broadcast, prif_co_reduce
+  use prif_test_m, only : test_description_substring
+  implicit none
+
+contains
+
+  impure elemental subroutine co_all(boolean)
+    logical(c_bool), intent(inout) :: boolean
+    call prif_co_reduce(boolean, c_funloc(both))
```
bonachea, line 14: This is using the PRIF 0.4 version of `prif_co_reduce`, and will break when we upgrade the implementation to PRIF 0.5, which significantly complicates the code needed to invoke `prif_co_reduce` (see PR #158).

Moreover there's no semantic reason to use a custom reduction function here. Please convert the boolean to an integer and use `prif_co_sum` or `prif_co_max` instead, to avoid the mess mentioned above.

test/prif_co_reduce_test_m.F90
bonachea, line 1: Changes in this file massively conflict with the complete rewrite in-progress in PR #158 to accommodate the PRIF 0.5 changes to collectives.

test/prif_co_max_test_m.F90
bonachea, line 1: Changes in this file will conflict with the changes in-progress in PR #158 to accommodate the PRIF 0.5 changes to collectives.

test/prif_co_min_test_m.F90
bonachea, line 1: Changes in this file will conflict with the changes in-progress in PR #158 to accommodate the PRIF 0.5 changes to collectives.

test/prif_co_sum_test_m.F90
bonachea, line 1: Changes in this file will conflict with the changes in-progress in PR #158 to accommodate the PRIF 0.5 changes to collectives.

test/prif_teams_test_m.F90
bonachea, line 1: Changes in this file massively conflict with the changes in-progress in PR #171, where this test was greatly expanded to complete coverage for the PRIF 0.5 implementation of teams.

bonachea (changes requested):
Made some additional observations. 

Right now my biggest concern with this PR is that it updates an (old?) snapshot of of the test suite, which is continuing to evolve in several other in-flight PRs. I don't want this PR to become a blocker for merging PRs that add Caffeine features (with corresponding test code that naturally will semantically conflict with this PR). 

Given a choice between reaching PRIF 0.5 compliance (a client-facing feature goal) and replacing an internal test harness, I'd currently lean towards prioritizing the former. Does the proposed switch from Veggies to Julienne provide any direct benefit to library clients, or is this purely a maintainability upgrade that mostly benefits Caffeine developers? 

Let's plan to discuss this further in our meeting this week.

bonachea:
The simple file renames for `test/*` proposed as part of this PR have now been merged in standalone PR #174, as previously discussed and agreed.

This PR will eventually need a rebase to resolve that conflict and also semantic conflicts with PR #171 and #175 that have also been merged. However as discussed work on this PR should probably be deferred until after PR #158 is completed and the test suite reaches a point of stability with PRIF 0.5, to avoid another redundant round of conflict resolution.

rouson:
This branch is out of date and Julienne has evolved with some breaking changes.  We can delete this branch and create a new PR after 
1. Julienne 2.0.0 releases and
2. Any other work that would touch the test suite is paused.

PR #170, opened 2025-01-03
Begin illustrating how to splice into veggies to work with prif

everythingfunctional:
This illustrates how you might only need to replace a handful of pieces of veggies to get it to play nicely with PRIF. All that's left from here is to copy over the remaining types of tests from veggies. and make the same tweaks to the `run*` procedures.

### DIFF
--- a/test/prif_veggies/prif_run_tests.f90
+++ b/test/prif_veggies/prif_run_tests.f90
@@ -1,0 +1,181 @@
+module prif_run_tests
+    use iso_c_binding, only: c_int64_t, c_size_t, c_ptr, c_null_ptr, c_null_funptr, c_f_pointer
+    use iso_fortran_env, only: int64, output_unit, error_unit
+    use iso_varying_string, only: put_line, var_str, operator(//)
+    use prif, only: &
+        prif_this_image_no_coarray, prif_num_images, &
+        prif_critical_type, prif_coarray_handle, &
+        prif_allocate_coarray, prif_deallocate_coarray, &
+        prif_co_reduce, prif_operation_wrapper_interface
+    use strff, only: to_string
+    use veggies, only: filter_item_result_t, test_item_t, test_result_item_t
+    use veggies_command_line_m, only: options_t, get_options, DEBUG
+
+    implicit none
+    private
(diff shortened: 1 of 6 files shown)


test/prif_veggies/prif_run_tests.f90
```diff
@@ -0,0 +1,181 @@
...
+                            // to_string(results%num_cases()) // " cases failed")
+                call put_line( &
+                        error_unit, &
+                        to_string(results%num_failing_asserts()) // " of " &
+                            // to_string(results%num_asserts()) // " assertions failed")
+                call put_line(error_unit, "")
+                suite_failed = .true.
+            end if
+
+            call prif_end_critical(critical_coarray)
```
bonachea, line 139: This scheme that attempts use of `CRITICAL` to isolate multi-line console output from different images is fundamentally flawed. In particular, `END CRITICAL` aka `prif_end_critical` do not guarantee that prior console output has been flushed all the way back to a spawning console that multiplexes the output from all images, and in general on distributed systems there is no way to guarantee that. That means the spawning console can still show interleaving of output from multiple images, despite the CRITICAL section used to guard output injection. 

The only 100% reliable way to ensure ordering of console output on any distributed system is to only perform output from a single image. This is also often the most scalable solution.

If we really want multi-line output from all images with "best effort" isolation, then in my experience the most reliable way to accomplish that "most of the time" is for each image to compose its entire output into a single long in-memory string with embedded newlines, and inject that output as a single operation (ideally using a single `write()` system call followed by `fsync()`). For rather long messages this can still become fragmented and potentially interleaved at the system level, but it's far more likely to be handled atomically by the output mulitplexor than a long series of independent output operations containing a single line each.

bonachea (commented):
I remain agnostic on the question of whether we should continue to use (modified) Veggies vs replace it with Julienne (PR #169) or something else.

However I added a few observations on the current code.

everythingfunctional:
As I pointed out on Slack, this extra complexity is now not needed to accomplish this. See how [garden](https://gitlab.com/everythingfunctional/garden) now does it.

PR #171, opened 2025-01-06
Add more team support 

### LINKED ISSUE #59: Add num_imgs and this_img as private components to `prif_team_type`
- [ ] Add `num_imgs` and `this_img` as private components to `prif_team_type`
- [ ] Change definitions of versions of `this_image` and `num_images` that do not use optional arguments to use the above components
- [ ] Add back in corresponding static global variables in `caffeine.c`

NOTE: All of the above are in reference to the initial state/ initial team and any identifiers should be named carefully to make this clear.

bonachea:
This PR implements:

* `prif_team_number`
* `prif_get_team`
* `prif_num_images_with_team`
* `prif_this_image_no_coarray(team)`

This notably completes implementation of the "Teams" chapter of PRIF 0.5, and makes significant progress towards completing "Image Queries".

This PR also fixes #59, thereby avoiding cross-language calls on every this_image and num_images query.

### DIFF
--- a/src/caffeine/caffeine.c
+++ b/src/caffeine/caffeine.c
@@ -28,11 +28,11 @@ typedef void(*final_func_ptr)(void*, size_t) ;
 typedef uint8_t byte;
 
 #if __GNUC__ >= 12
-  const int float_Complex_workaround = CFI_type_float_Complex;
-  const int double_Complex_workaround = CFI_type_double_Complex;
+  #define float_Complex_workaround  CFI_type_float_Complex
+  #define double_Complex_workaround CFI_type_double_Complex
 #else
-  const int float_Complex_workaround = 2052;
-  const int double_Complex_workaround =4100;
+  #define float_Complex_workaround  2052
+  #define double_Complex_workaround 4100
 #endif
 
 // ---------------------------------------------------
--- a/src/caffeine/caffeine.c
+++ b/src/caffeine/caffeine.c
@@ -35,11 +35,16 @@ typedef uint8_t byte;
   const int double_Complex_workaround =4100;
 #endif
 
-int caf_this_image(gex_TM_t team)
+// ---------------------------------------------------
+int caf_this_image(gex_TM_t gex_team)
 {
-  return gex_TM_QueryRank(team) + 1;
+  return gex_TM_QueryRank(gex_team) + 1;
 }
-
+int caf_num_images(gex_TM_t gex_team)
+{
+  return gex_TM_QuerySize(gex_team);
+}
+// ---------------------------------------------------
 // NOTE: gex_TM_T is a typedef to a C pointer, so the `gex_TM_t* initial_team` arg in the C signature matches the BIND(C) interface of an `intent(out)` arg of type `c_ptr` for the same argument
 void caf_caffeinate(
   mspace* symmetric_heap,
@@ -92,11 +97,6 @@ void caf_caffeinate(
 void caf_decaffeinate(int exit_code)
 {
   gasnet_exit(exit_code);
-}
-
-int caf_num_images(gex_TM_t team)
-{
-  return gex_TM_QuerySize(team);
 }
 
 void* caf_allocate(mspace heap, size_t bytes)
--- a/test/caf_teams_test.f90
+++ b/test/caf_teams_test.f90
@@ -1,16 +1,7 @@
 module caf_teams_test
-    use iso_c_binding, only: c_size_t, c_ptr, c_null_funptr, c_int64_t
-    use prif, only: &
-            prif_coarray_handle, &
-            prif_allocate_coarray, &
-            prif_deallocate_coarray, &
-            prif_this_image_no_coarray, &
-            prif_num_images, &
-            prif_team_type, &
-            prif_form_team, &
-            prif_change_team, &
-            prif_end_team
-    use veggies, only: result_t, test_item_t, assert_equals, describe, it, succeed
+    use iso_c_binding, only: c_size_t, c_ptr, c_null_funptr, c_int64_t, c_int
+    use prif
+    use veggies, only: result_t, test_item_t, assert_equals, assert_that, describe, it, succeed, fail
 
     implicit none
     private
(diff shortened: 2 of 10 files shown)


bonachea:
@ktras please review this PR in the next few days.

I now have a second Caffeine PR staged behind this one, I'm waiting for this one to be finalized and merged before posting it.

src/caffeine/caffeine.c
```diff
@@ -28,18 +28,23 @@ typedef void(*final_func_ptr)(void*, size_t) ;
 typedef uint8_t byte;
 
 #if __GNUC__ >= 12
-  const int float_Complex_workaround = CFI_type_float_Complex;
```
ktras, line 31: @bonachea Not asking for changes here, but just curious why you preferred for these to be macros?

test/caf_teams_test.f90
```diff
@@ -29,16 +20,62 @@ function check_teams() result(result_)
...
+        integer(c_int64_t) :: which_team, n
         integer, parameter :: num_coarrays = 4
         type(prif_coarray_handle) :: coarrays(num_coarrays)
         type(c_ptr) :: allocated_memory
-        type(prif_team_type) :: team
+        type(prif_team_type) :: team, initial_team, t
 
-        call prif_this_image_no_coarray(this_image=me)
         call prif_num_images(num_images=initial_num_imgs)
+        result_ = result_ .and. &
```
ktras, line 33: I think this first assignment to `result_` should probably not reference it on the right hand side as well.

src/caffeine/caffeine.c
bonachea: Great question!

This change was required to fix compile errors present in `main` when compiled with gcc 14.2 and `-O0`. 

Apparently with optimizations enabled, gcc is smart enough to constant propagate these `const` integer _variables_ to the `case` statement (which requires an "integer constant expression") and permissively accepts the old code that is not strictly C spec compliant.

Using a macro ensures the `{float,double}_Complex_workaround` tokens strictly qualify as "integer constant expression" for any C compiler, ensuring the portability of this code.

test/caf_teams_test.f90
```diff
@@ -29,16 +20,62 @@ function check_teams() result(result_)
...
+        call prif_get_team(level=PRIF_INITIAL_TEAM, team=t)
+        call prif_team_number(team=t, team_number=n)
+        result_ = result_ .and. &
+          assert_equals(int(n), -1, "prif_get_team(PRIF_INITIAL_TEAM) retrieves initial team")
+
+        t = prif_team_type() ; n = 0 ! clear outputs
+        call prif_get_team(level=PRIF_CURRENT_TEAM, team=t)
+        call prif_team_number(team=t, team_number=n)
+        result_ = result_ .and. &
+          assert_equals(int(n), -1, "prif_get_team(PRIF_CURRENT_TEAM) retrieves initial team")
```
ktras, line 71:
```suggestion
          assert_equals(int(n), -1, "prif_get_team(PRIF_CURRENT_TEAM) retrieves initial team when current team is initial team")
```

test/caf_teams_test.f90
```diff
@@ -29,16 +20,62 @@ function check_teams() result(result_)
...
+
+        t = prif_team_type() ; n = 0 ! clear outputs
+        call prif_get_team(level=PRIF_PARENT_TEAM, team=t)
+        call prif_team_number(team=t, team_number=n)
+        result_ = result_ .and. &
+          assert_equals(int(n), -1, "prif_get_team(PRIF_PARENT_TEAM) retrieves initial team")
```
ktras, line 77:
```suggestion
          assert_equals(int(n), -1, "prif_get_team(PRIF_PARENT_TEAM) retrieves initial team when parent team is initial team")
```

test/caf_teams_test.f90
```diff
@@ -48,6 +85,66 @@ function check_teams() result(result_)
...
+            call prif_get_team(level=PRIF_CURRENT_TEAM, team=t)
+            call prif_team_number(team=t, team_number=n)
+            result_ = result_ .and. &
+              assert_equals(int(n), int(which_team), "prif_get_team(PRIF_CURRENT_TEAM) retrieves current team")
+    
+            t = prif_team_type() ; n = 0 ! clear outputs
+            call prif_get_team(level=PRIF_PARENT_TEAM, team=t)
+            call prif_team_number(team=t, team_number=n)
+            result_ = result_ .and. &
+              assert_equals(int(n), -1, "prif_get_team(PRIF_PARENT_TEAM) retrieves initial team")
```
ktras, line 135:
```suggestion
              assert_equals(int(n), -1, "prif_get_team(PRIF_PARENT_TEAM) retrieves initial team when parent team is initial team")
```

test/caf_teams_test.f90
```diff
@@ -1,16 +1,7 @@
...
-            prif_deallocate_coarray, &
-            prif_this_image_no_coarray, &
-            prif_num_images, &
-            prif_team_type, &
-            prif_form_team, &
-            prif_change_team, &
-            prif_end_team
-    use veggies, only: result_t, test_item_t, assert_equals, describe, it, succeed
+    use iso_c_binding, only: c_size_t, c_ptr, c_null_funptr, c_int64_t, c_int
+    use prif
```
ktras, line 3: I think we should restore the `only-clause` for this `use-stmt`. We are using a lot of symbols from the module, but in general, we try to always use `only-clauses`. I personally don't need or prefer listing each one on its own line though, as this one did before it was changed.

src/caffeine/caffeine.c
ktras: Interesting, thanks for the explanation! Good to know!

test/caf_teams_test.f90
bonachea: Please explain the maintainability benefit of an explicit `only-clause` for a PRIF unit test. 

This test is designed to test a wide set of procedures provided by the PRIF module, as defined in the test code below. I see no benefit to redundantly listing them all at the top of the file, that just seems to create a maintainability headache while offering no real benefit.

To be clear, I do understand the general software engineering purpose of `only-clause`s as a way to avoid cross-module name conflicts. However the PRIF module's public exports have been carefully named to avoid such problems by design, i.e., we strictly follow the discipline of prefixing all `prif` module exports with `prif_` or `PRIF_`.

I want to ensure (and regression test) that clients can always safely `use prif` and be assured of not importing any symbol names that conflict with non-prefixed names. More specifically, if a blanket `use prif` directive ever imported a symbol that did not follow the PRIF naming convention and conflicted with a symbol defined by the test code, that would represent a defect in the PRIF implementation that we'd want to know about.

As such, I think this simpler version is superior in both maintainability and test coverage.

test/caf_teams_test.f90
bonachea: Veggies `result_t` is default-initialized to an empty result, so this is technically also correct as-written.

I want to maintain this pattern throughout the file, so I've added an initial assignment in f80e411b to address your readability concern.

bonachea:
@ktras thanks for your review, I believe I've addressed or responded to all of your concerns.

Ready for next round of review.

test/caf_teams_test.f90
ktras: You've convinced me that in this case the blanket policy of always using an `only-clause` is not gaining anything. And, as you've pointed out, it is taking away that testing opportunity for clashes. It certainly is more maintainable. I'm good with the code as is. Thanks for your response.

ktras (approved):
LGTM. Thanks!

PR #172, opened 2025-01-13
Use external assert library

bonachea:
This PR replaces Caffeine's internal use of the stub `prif:prif_private_s:caffeine_assert_s` submodule with the external [berkeleylab/assert](https://github.com/BerkeleyLab/assert) library. 

The main benefits of this upgrade:

1. **Macro-flavored assertions** that are guaranteed to compile away to zero runtime overhead when assertions are disabled, regardless of compiler analysis
2. **Automatic file/line reporting** to the console for failed assertions
3. **Broader type support for assert diagnostic data**, including `characterizable_t` and `intrinsic_array_t`
4. Leverage GASNet's automatic backtracing support and debugger integrations for assertion failures on any platform.

Here is example `fpm test` output from a manually injected assertion failure:
```
A total of 59 test cases

*** FATAL ERROR (proc 2): Assertion "prif_co_min: LHS/RHS length match in file ././src/caffeine/collective_subroutines/co_min_s.F90, line 30" failed on image 3 with diagnostic data "we        , like     "
*** NOTICE (proc 2): Before reporting bugs, run with GASNET_BACKTRACE=1 in the environment to generate a backtrace.
*** Caught a fatal signal (proc 2): SIGABRT(6)
<ERROR> Execution for object " caffeine-test " returned exit code  134
<ERROR> *cmd_run*:stopping due to failed executions
STOP 134
```

This PR notably utilizes the [`ASSERT_PARALLEL_CALLBACKS` feature](https://github.com/BerkeleyLab/assert/pull/51) added in assert library version 2.0.1, which was added specifically for Caffeine's use case. This ensures:

1. That a failing assertion reports the Caffeine-provided image number, rather than natively invoking the `THIS_IMAGE()` parallel feature.
2. That a failing assertion passes control back to Caffeine to handle job teardown, rather than natively invoking the `ERROR STOP` parallel feature.

Both of these are intended to break a dependency cycle, ensuring Caffeine's dependency on the assert library does not complicate builds for the compiler client by utilizing native parallel features, or result in an explicitly circular dependency back to the `prif` module.

Finally, this PR enables control of assertion enforcement on the install.sh command line, eg:
```
env CPPFLAGS=-DASSERTIONS=0 install.sh ...
```
the default remains assertions enabled while Caffeine development is still in a pre-production state.


rouson (approved):
LGTM

Thanks, @bonachea!

PR #176, opened 2025-01-16
Add CI that uses LLVM Flang to build Caffeine

ktras:


.github/workflows/build-with-flang.yml
```diff
@@ -0,0 +1,37 @@
...
+      run: |
+        fpm --version
+        $FC --version
+        $CC --version
+        export FPM_FC=$FC
+        export FPM_CC=$CC
+        export GASNET_PSHM_NODES=8
+        apt update
+        apt install -y build-essential pkg-config make
+        ./install.sh
```
bonachea, line 37: The step is called "Build and Test with LLVM Flang", but it doesn't actually do the latter.

Can we please an invocation of `./build/run-fpm.sh test -- -d`, ideally in a separate job step as in [CI.yml](https://github.com/BerkeleyLab/caffeine/blob/main/.github/workflows/CI.yml)?

.github/workflows/build-with-flang.yml
```diff
@@ -0,0 +1,37 @@
+name: Build with LLVM Flang
```
bonachea, line 1: Now that we have two compilers in CI, please rename the existing "CI" action in CI.yml, perhaps to "Build with gfortran"

bonachea (changes requested):
Very excited to see this just about ready!!

Made two requests

.github/workflows/build-with-flang.yml
ktras: Just added it, but not in a different step. I will make that change as requested.

bonachea: The separate step makes it much easier to reach the relevant output when something fails.

ktras: Yes, I agree. Great point and that has been done in the latest commit.

.github/workflows/build-with-flang.yml
ktras: Done. Thanks.

bonachea (approved):
LGTM!

PR #177, opened 2025-01-17
Resolve a minor TODO

bonachea:
This is a left-over from pull request #171 and #172 that was only possible to fully resolve now they are both merged.

### DIFF
--- a/src/caffeine/prif_private_s.F90
+++ b/src/caffeine/prif_private_s.F90
@@ -247,7 +247,6 @@ contains
     integer(c_int), intent(in) :: image_num
     integer(c_intptr_t), intent(out) :: ptr
 
-    integer(c_int) :: num_img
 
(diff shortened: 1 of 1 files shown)


src/caffeine/prif_private_s.F90
```diff
@@ -246,11 +249,7 @@ subroutine base_pointer(coarray_handle, image_num, ptr)
 
     integer(c_int) :: num_img
```
ktras, line 250:
```suggestion
```

ktras (changes requested):
Looks great, just one minor suggestion since I think `num_img` is now an unused variable.

src/caffeine/prif_private_s.F90
bonachea: Good catch!

ktras (approved):
LGTM

PR #184, opened 2025-03-12
Implement some low-hanging fruit

### LINKED ISSUE #113: Add implementation for `prif_size_bytes`
As mentioned by @bonachea in a PR review, implementation is potentially just:
`data_size = coarray_handle%info%coarray_size`

bonachea:
This PR plucks some "low-hanging fruit" from Caffeine's tree of unimplemented PRIF procedures.

Add fully compliant implementations for all of the following:

* `prif_{failed,stopped}_images`
* `prif_image_status`
* `prif_size_bytes`
* `prif_sync_team`
* `prif_sync_memory` 
* `prif_[lu]cobound*`
* `prif_coshape`

along with associated test code.

Make some other minor improvements.

Fixes #113.


src/caffeine/coarray_queries_s.F90
```diff
@@ -1,36 +1,56 @@
...
+    ucobounds = coarray_handle%info%ucobounds(1:coarray_handle%info%corank)
   end procedure
 
   module procedure prif_coshape
-    call unimplemented("prif_coshape")
+    type(prif_coarray_descriptor), pointer :: info
+
+    call_assert(coarray_handle_check(coarray_handle))
+
+    info => coarray_handle%info
```
ktras, line 43:
```suggestion
    associate(info => coarray_handle%info)
```

src/caffeine/coarray_queries_s.F90
```diff
@@ -1,36 +1,56 @@
...
+    sizes = info%ucobounds(1:info%corank) - info%lcobounds(1:info%corank) + 1
```
ktras, line 44:
```suggestion
      sizes = info%ucobounds(1:info%corank) - info%lcobounds(1:info%corank) + 1
    end associate
```

src/caffeine/coarray_queries_s.F90
```diff
@@ -1,36 +1,56 @@
...
   module procedure prif_ucobound_no_dim
-    call unimplemented("prif_ucobound_no_dim")
+    call_assert(coarray_handle_check(coarray_handle))
+
+    ucobounds = coarray_handle%info%ucobounds(1:coarray_handle%info%corank)
   end procedure
 
   module procedure prif_coshape
-    call unimplemented("prif_coshape")
+    type(prif_coarray_descriptor), pointer :: info
```
ktras, line 39:
```suggestion
```

docs/implementation-status.md
```diff
@@ -54,14 +54,14 @@ in the following sections.
...
+| `prif_num_images`                   | **YES** |  |
+| `prif_num_images_with_team`         | **YES** |  |
 | `prif_num_images_with_team_number`  | no |  |
-| `prif_this_image_no_coarray` | **YES** |  |
+| `prif_this_image_no_coarray`        | **YES** |  |
 | `prif_this_image_with_coarray`, `prif_this_image_with_dim`  | no |  |
-| `prif_failed_images` | no |  |
-| `prif_stopped_images` | no |  |
-| `prif_image_status` | no |  |
+| `prif_failed_images`                | **YES** |  |
```
ktras, line 62: Should these three have a note saying they are implemented but without support for catching failed or stopped images?

test/prif_coarray_inquiry_test.F90
```diff
@@ -49,4 +55,81 @@ function check_prif_local_data_pointer() result(result_)
...
+      integer(kind=c_int64_t), dimension(corank) :: lcobounds, ucobounds, tmp_bounds
+      integer(kind=c_int64_t) :: tmp_bound
+      integer(kind=c_size_t), dimension(corank)  :: sizes
+      type(prif_coarray_handle) :: coarray_handle
+      type(c_ptr) :: allocated_memory
+      integer(c_size_t) :: data_size, query_size
+  
+      result_ = succeed("")
+  
+      call prif_num_images(num_images=num_imgs)
```
ktras, line 76: @bonachea Can you help explain to me how we can get back a meaningful value to a `prif_num_images` call in a test if there wasn't previously a `prif_init` call in the test? I think I've been confused about this for a while for the testing context.

ktras (changes requested):
Looks great. I have some suggested changes and a question below.

src/caffeine/prif_private_s.F90
```diff
@@ -307,6 +317,25 @@ pure function optional_value(var) result(c_val)
...
+    type(prif_coarray_descriptor), pointer :: info
+    integer(c_int) :: i
+
+    call_assert(associated(coarray_handle%info))
+    info => coarray_handle%info
+    call_assert(info%corank >= 1)
+    call_assert(info%corank <= size(info%ucobounds))
+    call_assert(all([(info%lcobounds(i) <= info%ucobounds(i), i = 1, info%corank)]))
+    call_assert(info%coarray_size > 0)
+    call_assert(c_associated(info%coarray_data))
```
bonachea, line 334:
```suggestion
      call_assert(info%corank >= 1)
      call_assert(info%corank <= size(info%ucobounds))
      call_assert(all([(info%lcobounds(i) <= info%ucobounds(i), i = 1, info%corank)]))
      call_assert(info%coarray_size > 0)
      call_assert(c_associated(info%coarray_data))
```

docs/implementation-status.md
bonachea: I considered this but would strongly argue NO, on this basis:

These trivial implementations are nonetheless compliant with the letter of the F23  standard, which includes text like the following (emphasis added) for `STOPPED_IMAGES`:

> Result Value. The elements of the result are the values of the indices of the images that are _**known to have initiated normal termination**_ in the specified team, in numerically increasing order. **_If_** the executing image has previously executed an image control statement whose STAT= specifier assigned the value STAT_STOPPED_IMAGE from the intrinsic module ISO_FORTRAN_ENV or invoked a collective subroutine whose STAT argument was assigned STAT_STOPPED_IMAGE, at least one of the images participating in that image control statement or collective invocation shall be **_known_** to have initiated normal termination.

In other words, as long as no image control statement or collective ever defines `STAT=STAT_STOPPED_IMAGE`, then no images are guaranteed to be "known to have initiated normal termination" by the definition above. Caffeine never sets that STAT value, thus it's accurate to claim no image is ever guaranteed to be "known to have initiated normal termination", and thus a permitted/expected return value from `prif_stopped_images` is always an empty array. The same wording and logic applies to `FAILED_IMAGES` and `STAT_FAILED_IMAGE`.

Also note that `prif_fail_image` remains unimplemented (and might forever in Caffeine), so clients cannot even expect to generate a "user-injected" image failure on Caffeine. I wouldn't object to someday adding a caveat about limited failed-image support to some higher-level documentation about Caffeine, but I don't think it belongs at the PRIF entry-point level. Also it seems preliminary to start such a high-level list today given how many PRIF interfaces remain to be implemented.

test/prif_coarray_inquiry_test.F90
bonachea: That's a great question.

As stated in the PRIF spec:

> prif_init shall be called prior to any other calls to the PRIF implementation and shall be called at least once per process. Any second or subsequent call to prif_init by a given process is guaranteed to return immediately with no effect on system state, with PRIF_STAT_ALREADY_INIT assigned to the variable specified in the stat argument.

Here I'm relying on the call to `prif_init` that takes place inside the `a00_caffeinate_caffeinate()` unit test, which we've arranged with veggies to ensure it always runs first. All the other unit tests make the same assumption, which remains valid so long as all the unit tests run after `a00_caffeinate_caffeinate()` as part of the same job/process.

Note the codes in the `example/` directory (including the sub-jobs launched by the exit tests) are each independent `program`s, each of which runs in a separate job/process and thus those must include calls to `prif_init` (and `prif_stop`)

bonachea (commented):
Thanks for the review @ktras !

I've applied your suggestions and replied to the comment threads.

Ready for the next round of review.

docs/implementation-status.md
ktras: Thank you for the explanation. I agree with your choice.

src/caffeine/prif_private_s.F90
```diff
@@ -307,6 +317,25 @@ pure function optional_value(var) result(c_val)
     end if
   end function
 
+  ! verify state invariants for a coarray_handle
+  elemental impure function coarray_handle_check(coarray_handle) result(result_)
```
ktras, line 321:
```suggestion
  elemental function coarray_handle_check(coarray_handle) result(result_)
```
Since the pointer was taken out and the `assert` procedures are technically `pure`, this should work without the `impure` keyword. I tried it out earlier today with the `associate` and `elemental`.

test/prif_coarray_inquiry_test.F90
ktras: Awesome, thanks! That mystery is solved now for me!!

src/caffeine/prif_private_s.F90
bonachea: I did consider that, but I would like the flexibility to allow `coarray_handle_check` in a future version of Caffeine to eventually print more detailed diagnostic information to the console (ie I/O, which is not `pure`), so I'd prefer to reserve for that possibility via `impure` (at least until there's a compelling client for pure-ness here).

Also, the whole point of this procedure is sanity checking of in-memory state, so I _really_ don't want the fortran optimizer to play code-movement games with calls to this function.

ktras (approved):
LGTM

PR #187, opened 2025-03-13
install.sh: Fix the CLI for build/run-fpm.sh

bonachea:
Previously the CLI for build/run-fpm.sh was broken in multiple ways and would not accept command lines that were valid for the underlying fpm, which was confusingly the one providing help messages.

Add logic to the generated script to ensure we are not injecting invalid build options (e.g. `--profile`) for fpm subcommands that don't recognize such options.

Also, always echo the underlying fpm command to be more transparent about what we're actually invoking, to help the user understand what's happening when things go wrong.


rouson (approved):
LGTM

PR #189, opened 2025-03-19
CI: Add run of some examples

bonachea:
This ensures our hello example gets coverage, and also establishes whether basic Caffeine functionality is working, separately from the veggies test harness.

Also separately run a few of the stop unit tests to validate exit code propagation.


ktras (approved):
LGTM

PR #190, opened 2025-03-19
fix #188: check for git in install.sh

### LINKED ISSUE #88: Implement graceful error handling for allocation failures
In the near term, Caffeine will likely treat _most_ errors as immediately fatal (ideally with a high-quality message as part of the error crash).

However one particularly important error that IMO should not get this treatment is memory allocation. Unlike hardware failure, out-of-memory is a common condition when scaling problems in real production science, and needs to be handled in a robust manner by a production-quality runtime. It's even plausible that some applications might perform non-trivial recovery from allocation failure.

`prif_allocate` and `prif_allocate_non_symmetric` currently ignore the possibility of errors and I suspect they crash in obscure ways upon memory exhaustion. IMO these two calls should be fixed to strictly adhere to Fortran error handling semantics, specifically wrt returning meaningful `stat` and `errmsg` (when provided) or crashing with a useful console message (when not provided). Ideally the error message in either case should include status information about the initial and current state of the shared heaps, and recommendations to the end-user about how to resolve the problem.

### LINKED ISSUE #188: Caffeine installation process using install.sh
I am using ```install.sh``` actually to install CAFFEINE as part of the JOSS review process (openjournals/joss-reviews#7895) using a (fresh) Ubuntu 24.04.2 LTS system. Here are my findings:

As already explained with CAFFEINE issue https://github.com/BerkeleyLab/caffeine/issues/97 using CAFFEINE ```install.sh``` to install **fpm** does not work. The issue with installing **fpm** is still there, problem description and solution can be found here:
https://fortran-lang.discourse.group/t/installing-fpm-0-10-1/8268/5

I did use the following to install **fpm** successfully manually:
```
git clone https://github.com/fortran-lang/fpm.git
cd fpm
sh ./install.sh
```

I did also install **Homebrew** manually because CAFFEINE ```install.sh``` does not install **curl** and **git**. 

### Here is my complete successful CAFFEINE installation process:

1. Install **Homebrew** manually as it is explained here: https://linuxconfig.org/installing-homebrew-on-ubuntu-debian-a-step-by-step-guide .

2. Run the CAFFEINE ```install.sh``` to install the compilers.

3. Install **fpm** manually as explained:
```
git clone https://github.com/fortran-lang/fpm.git
cd fpm
sh ./install.sh
```
and add **fpm** to your ```PATH```.

4. Open a new terminal window and re-run the CAFFEINE ```install.sh``` to successfully complete the installation process.

rouson:
If the `install.sh` script can't execute `git`, the script will now print a message early requesting the user install `git`, after which the installer exits with non-zero status.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -52,6 +52,7 @@ the latest versions using Homebrew:
   GCC $GCC_VERSION
   GASNet-EX $GASNET_VERSION
   fpm
+  git (used by fpm to clone dependencies)
   pkg-config
   realpath (Homebrew coreutils)
   GNU Make (Homebrew coreutils)
@@ -118,6 +119,12 @@ fi
 
 if command -v fpm > /dev/null 2>&1; then
   FPM=`which fpm`
+fi
+
+if ! command -v git > /dev/null 2>&1; then
+  echo "git not found. Building Caffeine requires fpm, which uses git to install dependencies."
+  echo "Please install git, ensur it is in your PATH, and rerun ./install.sh"
+  exit 1
 fi
 
 ask_permission_to_use_homebrew()
@@ -198,7 +205,7 @@ if [ -z ${FC+x} ] || [ -z ${CC+x} ] || [ -z ${CXX+x} ] || [ -z ${PKG_CONFIG+x} ]
     exit_if_user_declines "brew"
 
     if ! command -v curl > /dev/null 2>&1; then
-      echo "No download mechanism found. Please install curl and rerun ./install.sh"
+      echo "curl not found. Please install curl, ensure it is in your PATH, and rerun ./install.sh"
       exit 1
     fi
 
--- a/install.sh
+++ b/install.sh
@@ -53,6 +53,7 @@ the latest versions using Homebrew:
   GASNet-EX $GASNET_VERSION
   fpm
   git (used by fpm to clone dependencies)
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -198,7 +205,7 @@ if [ -z ${FC+x} ] || [ -z ${CC+x} ] || [ -z ${CXX+x} ] || [ -z ${PKG_CONFIG+x} ]
     exit_if_user_declines "brew"
 
     if ! command -v curl > /dev/null 2>&1; then
-      echo "No download mechanism found. Please install curl and rerun ./install.sh"
+      echo "curl not found. Please install curl, ensure it is in your PATH, and rerun ./install.sh"
       exit 1
     fi
```
bonachea, line 210: We agreed in our meeting that (just like `git`) `curl` is also always required (due to GASNet), so this entire block should also be promoted out of this HomeBrew branch up to an earlier point where it always runs (probably right beside where you added the `git` check).

Similarly `curl` should also be added to the dependency message.

install.sh
bonachea, line 1: git and curl should also both be added to the README list as build dependencies

bonachea (changes requested):
Good start on addressing issue #188.

Just a few remaining resolutions from our consensus.

PR #193, opened 2025-03-24
(chore): Pre-release chores

ktras:


bonachea:
@ktras @rouson What do we think about addressing #164 and #166 as part of this release cycle?

Seems silly to keep packaging files for (and wasting workflow minutes on) creating FORD docs that we are not using, don't link and do not recommend anyone read.

I'd really like to see us jettison all this dead FORD infrastructure.

bonachea (changes requested):
I think this mostly addresses step 4 in the [release procedure](https://github.com/BerkeleyLab/caffeine/blob/main/docs/README-release.md)

Added a few minor requests, but otherwise looks good!

ktras:
> @ktras @rouson What do we think about addressing #164 and #166 as part of this release cycle?
>

I believe Damian will be on a plane within minutes for work travel, so I don't know when he will be able to give his feedback on this. Based on our brief discussions in the past, I remember being on board with removing Ford docs from the repo and the CI, but since we didn't make a definitive team decision about it at the time, I think we need to discuss it as a group again. With the passage of time, I can't remember if there was a counter argument to the idea that it might be helpful to keep it because it might help potential contributors to Caffeine better understand the structure of the code and figure out how to get involved. However, these people could obviously follow a Ford command and produce the Ford docs locally.

If we want to keep the original deadline of producing a new release this Wednesday, we might want to hold off on any action items related to these issues until after the release. If so, we can still do the action items related to the resolution of these issues soon after the release once we come to a team decision.

If we want to delay the release a couple days until after our Wednesday meeting, we could try to make the team decision during that meeting and then do any action items before producing a new release later this week.

bonachea:
> I believe Damian will be on a plane within minutes for work travel, so I don't know when he will be able to give his feedback on this. Based on our brief discussions in the past, I remember being on board with removing Ford docs from the repo and the CI, but since we didn't make a definitive team decision about it at the time, I think we need to discuss it as a group again.

I've opened PR #194 to independently pursue the suggested removal. 
If everyone approves before the end of our Wednesday meeting then it can become part of the release ;-)

bonachea (approved):
Everything LGTM

PR #194, opened 2025-03-25
Remove automated invocation of FORD

### LINKED ISSUE #164: Investigate if `docs/index.md` is needed

### LINKED ISSUE #166: Decide if FORD docs need to remain in CI and repo

bonachea:
The PRIF specification: https://go.lbl.gov/prif
is the best and only reliable documentation for clients of the PRIF interface exported by Caffeine.

The documentation generated by FORD from the repo lacks all semantic information, and provides no clear-cut distinction between what is part of the stable PRIF interface and what is an internal Caffeine implementation detail.

As such, the FORD docs do more harm than good, so stop generating them automatically.

Fixes #164 
Fixes #166


rouson:
For every other project, I would very much oppose removing the FORD.  It's especially useful for the dependency graphs, which can save a lot of time in the early phases of pruning thousands of lines down to s 25-line reproducer.  Being able to see the internal dependencies at a glance saves a lot of time. It's also nice be able to see certain groupings at a glance such as as all the derived types in one place.

With all that said, Caffeine is the one exceptional case. We have a detailed specification document that's the best form of documentation. Also, Caffeine hasn't generated any compiler bug reports in recent times. And the target audience is compiler developers who presumably would prefer a more formal spec over the more informal documentation generated from comments in the code.

rouson:
I meant to review and approve but misclicked and only left a comment.  Now I don't see how to do an actual review -- possibly because I'm using the GitHub mobile app.

ktras (approved):
LGTM. I have some more thoughts about the FORD docs, but I think this is the correct move and I can share my other thoughts in a meeting.

PR #199, opened 2025-03-26
CI: Factor, combine and expand our GitHub actions

bonachea:
The Build Action now expands to jobs covering:

* gfortran-{13,14} on ubuntu-24.04, macos-13-x86_64, macos-14-arm64
* flang-new (currently v20 git snapshot) on ubuntu-24.04

In addition, the action recipe is now fully factored, simplifying future maintenance.


.github/workflows/build.yml
```diff
@@ -0,0 +1,112 @@
...
+        sudo apt install -y build-essential gfortran-${GCC_VERSION} g++-${GCC_VERSION} pkg-config make
+
+    - name: Install Ubuntu Container Dependencies
+      if: ${{ contains(matrix.os, 'ubuntu') && matrix.container != '' }}
+      run: |
+        set -x
+        apt update
+        apt install -y build-essential pkg-config make
+
+    - name: Install macOS Dependencies
```
ktras, line 57: When looking at the actions, it looks like this step is not run of any of the macOS instances. I'm not sure why. The only thing we are losing though is the `set -x` of course. But that is being set in other steps anyway, so perhaps this `Install macOS Dependencies` step should go away?

ktras:
@bonachea Please see my comment above and let me know what you think. Otherwise, the PR looks great. Also, once I approve the PR, you will need to go and change some settings, I think. Because the PR doesn't want to allow merging until the 3 CI actions pass that are on `main`, but that aren't on this branch, so they never launch.

.github/workflows/build.yml
ktras: I can see now why the step is not being run in the next line of code down from this commented line of code. So I am assuming this was deliberate to have this step in the file while not running it. Could you just explain to me why?

bonachea: Correct - I was just leaving a placeholder where we can easily add dependency commands in the future if/when that becomes necessary (at which point we'd re-enable the step of course).

If you think this is too confusing I can remove the `&& 0` to always run the empty step for macos

ktras: No, I don't think people will be looking at it that closely to be confused, I just was because of needing to review the PR. But it being a purposeful placeholder is a good enough reason to keep it as is. I just wanted to make sure it wasn't accidentally included.

ktras (approved):
LGTM. Feel free to change the repo settings as needed to be able to merge the PR even though the 3 actions from the `main` branch CI can't run.

PR #201, opened 2025-03-27
Add install.sh --verbose

bonachea:
Add an `install.sh --verbose` option and deploy it in CI for more useful logging


ktras (approved):
LGTM

Issue #157, opened 2024-12-19
Update PRIF_VERSION_MINOR to 0.5


bonachea:
This should be the last step, after we've made all the other interface changes for PRIF 0.5

bonachea:
Resolved in #207

PR #207, opened 2025-04-05
Update PRIF_VERSION_MINOR to 0.5

bonachea:
PR #158 brought the collectives interface into compliance with PRIF 0.5.

As such the prif.F90 provided by Caffeine is now believed to contain correct interfaces for every entry point specified in PRIF 0.5, and all are backed by linkable procedures.

Some procedures still throw an unimplemented error if invoked at runtime (see docs/implementation-status.md for the latest on implementation status), but the interfaces are complete.

Resolves issue #157


ktras (approved):
LGTM

PR #213, opened 2025-04-21
Hoist all use statements out of sub-submodules

bonachea:
use statements in a sub-submodule trigger pathological use conflicts in some compilers (gfortran) creating a maintenance headache.

Avoid this by hoisting all such use statements to the intermediate prif_private_s submodule.


ktras (approved):
LGTM

Issue #228, opened 2025-05-08
Support Homebrew-installed LLVM 20 on macOS


rouson:
On macOS 15.4.1 with a Homebrew-installed LLVM 20.1.4, `flang-new --version` yields
```
Homebrew flang version 20.1.4
Target: arm64-apple-darwin24.4.0
Thread model: posix
InstalledDir: /opt/homebrew/Cellar/flang/20.1.4/libexec
Configuration file: /opt/homebrew/Cellar/flang/20.1.4/libexec/flang.cfg
Configuration file: /opt/homebrew/etc/clang/arm64-apple-darwin24.cfg
```
which isn't recognized by Caffeine's `install.sh` script.  I just created PR #226 to address this.  After fixing that issue, the next error that I encounter is
```
 exec /Users/rouson/.local/bin/fpm build --profile debug --flag '-g -O3 -DASSERT_MULTI_IMAGE -DASSERT_PARALLEL_CALLBACKS -DASSERTIONS  -O3 -g' --compiler /opt/homebrew/Cellar/flang/20.1.4/libexec/flang-new --c-compiler /opt/homebrew/Cellar/llvm/20.1.4/bin/clang --c-flag '-O3 -Wno-unused -Wunused-result -Wno-unused-parameter -Wno-address  -D_GNU_SOURCE=1 -DGASNET_SEQ -I/Users/rouson/.local/include -I/Users/rouson/.local/include/smp-conduit' --link-flag '-O3 -Wno-unused -Wunused-result -Wno-unused-parameter -Wno-address  -L/Users/rouson/.local/lib'
[  0%]                     caffeine.c
[  2%]                     caffeine.c  done.

././src/caffeine/caffeine.c:15:10: fatal error: 'ISO_Fortran_binding.h' file not found
   15 | #include <ISO_Fortran_binding.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
<ERROR> Compilation failed for object " src_caffeine_caffeine.c.o "
<ERROR> stopping due to failed compilation
STOP 1
```
From what I can tell, I have two copies of `ISO_Fortran_binding.h` installed with LLVM 20:
```
/opt/homebrew/Cellar/flang/20.1.4/include/flang/ISO_Fortran_binding.h
/opt/homebrew/Cellar/flang/20.1.4/lib/clang/20/include/ISO_Fortran_binding.h
```
@ktras @bonachea  any thoughts on why it's not being found?

[...]

PR #231, opened 2025-05-08
doc(README): add warning about brew-installed LLVM

rouson:


README.md
```diff
@@ -46,6 +46,8 @@ Prerequisites & Dependencies
 The `install.sh` script uses the following packages:
 * Fortran and C compilers
     * We regularly test with: gfortran v14 and LLVM Flang 20
+    * **Note:** Using the Homebrew-installed llvm and flang packages to build
+      Caffeine fails with an error related to not finding "ISO_Fortran_binding.h".
```
bonachea, line 50: How confident are we this problem is universal and not specific to one test system?

I suggest we soften this a bit and reference the open issue:

```suggestion
    * We've encountered some problems on macOS using the Homebrew-installed llvm and flang 
      packages, see issue #228 for the latest information.
```

bonachea (changes requested):

README.md
rouson: @ktras please see if you can reproduce this behavior on your system.

@bonachea I suspect it's related to the fact that Homebrew installs "llvm" and "flang" as separate packages.  The former packages installs `clang` and other software.  The latter of course installs `flang`.  For some reason, they get installed in different paths and "ISO_Fortran_binding.h" lives in a strange world wherein it is defined by the Fortran standard but exists only for purposes of serving C (notwithstanding ISO_Fortran_binding_m) so I'm not surprised the `clang` doesn't know where to find "ISO_Fortran_binding.h" --- especially given that the "flang" package is a lot newer and didn't even exist before a year or two ago as far as I know.

bonachea: @rouson That sounds like a plausible explanation that is worth copying into the issue tracker so that others can find it again later.

Generally when adding in-repo documentation for a known defect, I find it's best to keep the in-repo description minimal (just enough detail to be recognizable) and then provide a reference to an online issue with the full and most up-to-date details. This is especially true for defects that seem to involve external software (in this case, Homebrew). The issue in the online tracker may continue to evolve after a Caffeine release, for example with new workarounds or documenting versions of involved software where the defect was resolved.

For all these reasons I stand by my suggested change.

bonachea: As of PR #232 we now have macOS Homebrew flang in CI, using the workaround documented in issue #228. That workaround is functional today, but could potentially need adjustment in response to future Homebrew formulae updates, so I don't think we should hard-code it into our scripts.

@rouson I propose that we apply my suggested change above to the README text, and then this matter can be closed.

rouson: @bonachea I just committed your suggested revision.  Then I clicked "re-review" for you before reading your note above indicating that the matter can be closed after making the suggested change. For the paper trail, it probably makes sense for you to click approve before this is merged.  Feel free to merge it after doing so and after the tests pass.

bonachea (approved):

PR #232, opened 2025-05-09
CI: Add macos-15 and macos + flang20 (Homebrew)

bonachea:


.github/workflows/build.yml
```diff
@@ -73,29 +90,33 @@ jobs:
...
         set -x
-        if test "$FC" = "flang" ; then \
-          echo "FC=flang-new" >> "$GITHUB_ENV" ; \
-        else \
-          echo "FC=gfortran-${GCC_VERSION}" >> "$GITHUB_ENV" ; \
-        fi
-        if test "$CC" = "gcc" ; then echo "CC=gcc-${GCC_VERSION}" >> "$GITHUB_ENV" ; fi
-        if test "$CXX" = "g++" ; then echo "CXX=g++-${GCC_VERSION}" >> "$GITHUB_ENV" ; fi
+        brew update
+        brew install llvm@${GCC_VERSION} flang
```
rouson, line 97: Why is `GCC_VERSION` used in `llvm@${GCC_VERSION}`?  Should there be a defined `LLVM_VERSION that is used here instead?

bonachea: That was just me being lazy and re-using the variable, but I can switch things up to clarify

rouson: Fascinating. I'm surprised it works. Yes, I'd recommend switching it.

bonachea: Resolved in 5a3cba49

bonachea:
Apparently some kind of intermittent outage just began at gitlab.com, which is causing `fpm` to fail when fetching Veggies in CI [example](https://github.com/BerkeleyLab/caffeine/actions/runs/14940599011/job/41977403036?pr=232)

Even trying to load the [Veggies web page](https://gitlab.com/everythingfunctional/veggies) I'm seeing intermittent HTTP 500 failures, so I'm confident this has nothing to do with our CI.

image

bonachea:
https://status.gitlab.com just noticed/acknowledged the problem

image

rouson (approved):
LGTM

PR #242, opened 2025-07-11
CI: Tell setup-fpm to use the latest fpm

bonachea:
This ensures the setup-fpm action in CI uses the latest fpm (currently 0.12.0), instead of the default of 0.11.0


ktras (approved):
LGTM

Issue #245, opened 2025-08-13
Loss of stdout for prif_error_stop and quiet prif_stop on distributed conduits


bonachea:
Testing with Caffeine and flang 21 on distributed conduits (udp and ofi) reveals that we are sometimes losing final output from the caller that might be buffered in the Fortran-level `output_unit`, most commonly in jobs that end in `prif_error_stop` or `prif_stop(quiet=.true.)`.

PR #246, opened 2025-08-13
issue #245: prif_(error)_stop: Work harder to flush final output

bonachea:
Previously Caffeine was not always explicitly flushing Fortran stdout/stderr buffers during calls to `prif_stop` or `prif_error_stop`, leading to observed loss of final `output_unit` output (from prior write commands in the Fortran program) on some compilers and systems.

Change `prif_stop` and `prif_error_stop` to always flush Fortran-level `error_unit` and `output_unit` buffers, both immediately upon entry and again before the call to caf_decaffeinate.

Fixes issue #245

### DIFF
--- a/src/caffeine/program_termination_s.F90
+++ b/src/caffeine/program_termination_s.F90
@@ -14,6 +14,12 @@ submodule(prif:prif_private_s) program_termination_s
 
 contains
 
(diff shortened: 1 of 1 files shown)


ktras (approved):
LGTM

PR #260, opened 2025-09-19
Fix issue #259: prif_init signals IEEE_INEXACT

### LINKED ISSUE #259: prif_init signals IEEE_INEXACT
When running the new `app/native-multi-image.F90` test with flang-latest, both [in CI](https://github.com/BerkeleyLab/caffeine/actions/runs/17847426037/job/50749540383) and manually on Perlmutter, each image generates output at `STOP` that looks like this:

```
Fortran STOP
IEEE arithmetic exceptions signaled: INEXACT
```

The messages are output by flang's implementation of the `STOP` statement, which is reporting the signalling state of IEEE exceptions to the console (as required by the F23 standard).

The IEEE exception appears to be arising from within Caffeine's `prif_init`, somewhere within the `BIND(C)` call to `caf_caffeinate`. The effect is that any program compiled `flang -fcoarray` that doesn't otherwise clear the IEEE exceptions will currently generate this message upon reaching a non-quiet native `STOP`.

bonachea:
Fixes #259

### DIFF
--- a/src/caffeine/program_startup_s.F90
+++ b/src/caffeine/program_startup_s.F90
@@ -7,6 +7,7 @@ submodule(prif:prif_private_s) program_startup_s
 contains
 
   module procedure prif_init
+    use ieee_arithmetic, only: ieee_inexact, ieee_set_flag
     logical, save :: prif_init_called_previously = .false.
 
     if (prif_init_called_previously) then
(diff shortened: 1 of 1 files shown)


ktras (approved):
LGTM

PR #269, opened 2025-11-11
Add team features to native-multi-image.F90 app

ktras:
This PR will not pass CI (with the current macros being set for flang) until [PR #165573](https://github.com/llvm/llvm-project/pull/165573) is merged into the llvm-project repo.

### DIFF
--- a/app/native-multi-image.F90
+++ b/app/native-multi-image.F90
@@ -115,11 +115,7 @@ program native_multi_image
   res = GET_TEAM()
   write(*,'(A,I3)') "Initial team number is ", TEAM_NUMBER()
 
-  if (ni == 1) then
-    team_id = ni
-  else
-    team_id = merge(1, 2, me <= ni/2)
-  end if
+    team_id = merge(1, 2, me <= (ni+1)/2)
 
   FORM TEAM(team_id, subteam)
   SYNC TEAM(subteam)


ktras:
@bonachea I have not used macros for each individual team feature. In my initial draft of adding macros to my changes to this program, I did use macros for each individual team feature. However, particularly because of `team_number`, the use of macros around each different team number was a lot. I didn't prefer that style, so I ended up with the current strategy. If you feel strongly about the additional macros for each separate team feature, I can update the PR to add them back.

app/native-multi-image.F90
```diff
@@ -34,11 +34,15 @@ program native_multi_image
...
+#endif
+#ifndef HAVE_TEAM
+#define HAVE_TEAM 1
 #endif
 
   USE, INTRINSIC :: ISO_FORTRAN_ENV
-  integer :: me, ni, peer, tmp
+  integer :: me, ni, peer, tmp, team_id
   character(len=5) :: c
+  type(team_type) :: subteam, res
```
bonachea: This declaration should be conditional on `HAVE_TEAM`

bonachea (changes requested):
Recording one new observation

app/native-multi-image.F90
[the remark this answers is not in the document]
bonachea: In #273 I've just augmented our CI to detect these unwanted exit-time exceptions.

I've also tracked down the source of the problem in this particular case which is arising in `prif_form_team`. It should be fixed by commit 71e87f4e I just added to this branch.

app/native-multi-image.F90
```diff
@@ -91,6 +95,27 @@ program native_multi_image
...
+#if HAVE_TEAM
+  call status("Testing TEAMS...")
+  res = get_team(CURRENT_TEAM)
+  res = get_team(INITIAL_TEAM)
+  res = get_team()
+  write(*,'(A,I3)') "Initial team number is ", team_number()
+
+  if (ni < 2) then
+    if (me == 1) write(*,'(A)') "Please run program again with at least 2 images to test more TEAM features."
+  else
```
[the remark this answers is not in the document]
ktras: I agree, there isn't a good reason not to do it for one image. I have updated the test accordingly.

app/native-multi-image.F90
```diff
@@ -34,18 +34,37 @@ program native_multi_image
...
 #endif
 
   USE, INTRINSIC :: ISO_FORTRAN_ENV
   integer :: me, ni, peer, tmp
+# if HAVE_TEAM
+  integer :: team_id
+# endif
   character(len=5) :: c
+# if HAVE_TEAM
+  type(team_type) :: subteam, res
```
bonachea: Minor simplification:
```suggestion
  character(len=5) :: c
# if HAVE_TEAM
  integer :: team_id
  type(TEAM_TYPE) :: subteam, res
```

app/native-multi-image.F90
```diff
@@ -91,6 +110,28 @@ program native_multi_image
...
+  res = GET_TEAM(CURRENT_TEAM)
+  res = GET_TEAM(INITIAL_TEAM)
+  res = GET_TEAM()
+  write(*,'(A,I3)') "Initial team number is ", TEAM_NUMBER()
+
+  if (ni == 1) then
+    team_id = ni
+  else
+    team_id = merge(1, 2, me <= ni/2)
+  end if
```
bonachea: If there's only one image, it doesn't matter which team_id it uses

```suggestion
  team_id = merge(1, 2, me <= ni/2)
```

bonachea: OK in that case how about:

```suggestion
    team_id = merge(1, 2, me <= (ni+1)/2)
```

PR #280, opened 2025-12-19
issue #279: Move run-fpm.sh and fpm.toml out of build directory

### LINKED ISSUE #279: install.sh should not generate files into `./build`
`fpm` assumes it fully owns the contents of the `build` directory, and in particular the entire `build` directory is erased by `fpm clean --all`. Therefore the `build` directory should only contain files that `fpm` is capable of re-creating itself.

The two files generated by `install.sh` (`fpm.toml` and `run-fpm.sh`) cannot be re-created by `fpm`, thus do not belong in `build`.

bonachea:
Fixes #279

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,9 @@
 # fpm-generated build tree
 build
 
-# install.sh-generated fpm manifest 
+# install.sh-generated fpm files
 fpm.toml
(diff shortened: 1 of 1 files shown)


ktras (approved):
LGTM

PR #292, opened 2026-01-30
Improve error condition handling

### LINKED ISSUE #87: Ensure all PRIF calls with an optional stat argument actually set it when present
Most of the PRIF calls accept an optional `stat` argument, and many/most are currently completely ignoring it which is non-compliant.
We should audit all PRIF entry points to ensure they include code like :
```
if (present(stat)) stat=0
```

### LINKED ISSUE #128: Add proper handling of shared memory exhaustion
Currently Caffeine always crashes with obscure signals on shared memory exhaustion. Shared heap exhaustion is a very common failure mode in practice and probably the most important runtime error to detect and issue a high-quality diagnostic.

At a minimum, Caffeine needs to be updated to be PRIF-compliant and return `stat=PRIF_STAT_OUT_OF_MEMORY` when `stat` is provided.

I'd also like to see us format a nice error message that includes the total size of the current shared heap, the total size of the primordial shared heaps and the size of the request that failed. Then upon failure we can either return that message in `errmsg` (if provided) or print it to the console along with error termination.

bonachea:
Fixes #87
Fixes #128

Example out-of-memory message:
```
ERROR STOP 'Fortran shared heap is out of memory
  while allocating 500 TiB of additional coarray memory.

  Shared heap size information:
    Total shared heap:               1 GiB    (CAF_HEAP_SIZE)
    Total non-coarray heap:        102 MiB    (CAF_COMP_FRAC * CAF_HEAP_SIZE)
    Current team coarray heap:     921 MiB

  Consider setting the CAF_HEAP_SIZE environment variable to request a larger heap.'
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -289,7 +289,7 @@ jobs:
           CAF_IMAGES=$(( CAF_IMAGES / 2 )) ; \
         done
 
-    - name: Run exit tests
+    - name: Run exit/failure tests
       run: |
         echo CAF_IMAGES=${CAF_IMAGES}
         set -x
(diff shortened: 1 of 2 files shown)


ktras (approved):
LGTM and I have verified that all of the PRIF procedures that have a `stat` argument are now dealt with.

PR #307, opened 2026-03-19
team_check: coarrays

bonachea:
A minor follow-on enhancement to #306 : Expand `team_check` to also validate data structure invariants on the `coarrays` field.


ktras (approved):
LGTM

PR #320, opened 2026-05-05
Overhaul coarray finalization, convert `prif_coarray_handle` to be interoperable 

bonachea:
Implement the following changes expected in PRIF 0.8, as proposed in https://github.com/BerkeleyLab/prif/issues/157 (proposals 2, 2.1. 2.2, 2.3) and approved in https://github.com/BerkeleyLab/prif/pull/159 :

* Convert `prif_coarray_handle` to an interoperable type
* Convert the `final_func` argument to `prif_allocate_coarray` into a procedure pointer named `final_proc`, 
  while adjusting the callback interface and removing the `stat`/`errmsg` arguments.
* Convert `prif_local_data_pointer`, `prif_size_bytes` and `prif_{get,set}_context_data` to be interoperable


bonachea:
This is a somewhat large PR, primarily due to the pervasive impact of changing the internal representation of `prif_coarray_handle`, and (to a lesser extent) due to the breaking changes in the finalization interface. 

I recommend the following reading order for most logical review:

1. src/prif.F90
2. src/caffeine/unit_test_parameters_m.F90
3. test/test-uses-alloc.F90
4. test/prif_allocate_test.F90
5. test/prif_support_test.c
6. other test code
7. src/caffeine/prif_private_s.F90
8. src/caffeine/allocation_s.F90
9. everything else

test/prif_allocate_test.F90
```diff
@@ -23,15 +23,24 @@ module prif_allocate_test_m
...
   type(test_diagnosis_t) :: ff_diag
+#if CAF_PRIF_VERSION < 8
   logical :: ff_force_fail = .false.
   character(len=*), parameter :: ff_err = "test error message"
 #endif
 
+#if CAF_PRIF_VERSION >= 8
+  interface
+    subroutine coarray_cleanup_simple_c(handle) bind(C)
+      import c_int, c_char, prif_coarray_handle
```
ktras:
```suggestion
      import prif_coarray_handle
```

example/support-test/out_of_memory.F90
```diff
@@ -1,5 +1,7 @@
 program out_of_memory
-  use iso_c_binding, only: c_bool, c_size_t, c_ptr, c_null_funptr, c_int64_t
+# include "../../test/test-uses-alloc.F90"
```
ktras: This file changes here don't seem to indicate a need to include this file. I'm not seeing any usage of the `final_proc` function like macro. Can this include line be removed?

ktras (changes requested):
I have only 3 files left to review, but I have to put this down until tomorrow, so here is an initial review where everything looks great so far but please see these two minor comments.

example/support-test/out_of_memory.F90
bonachea: Agreed - good catch!
```suggestion
```

ktras (approved):
LGTM.

PR #337, opened 2026-06-24
Add more multi image intrinsic function calls to the integration

ktras:
test.

### DIFF
--- a/app/native-multi-image.F90
+++ b/app/native-multi-image.F90
@@ -58,13 +58,30 @@ program native_multi_image
 #endif
 
 #ifndef HAVE_COARRAY
-#define HAVE_COARRAY 0
+#define HAVE_COARRAY 1
 #endif
 #ifndef HAVE_MAIN_COARRAY
 #define HAVE_MAIN_COARRAY HAVE_COARRAY
 #endif
 #ifndef HAVE_ALLOC_COARRAY
 #define HAVE_ALLOC_COARRAY HAVE_COARRAY
+#endif
+
+! coarray query intrinsics
+#ifndef HAVE_COARRAY_QUERY
+#define HAVE_COARRAY_QUERY HAVE_COARRAY
+#endif
+#ifndef HAVE_COBOUND
+#define HAVE_COBOUND HAVE_COARRAY_QUERY
+#endif
+#ifndef HAVE_COSHAPE
+#define HAVE_COSHAPE HAVE_COARRAY_QUERY
+#endif
+#ifndef HAVE_IMAGE_INDEX
+#define HAVE_IMAGE_INDEX HAVE_COARRAY_QUERY
+#endif
+#ifndef HAVE_THIS_IMAGE_COARRAY
+#define HAVE_THIS_IMAGE_COARRAY HAVE_COARRAY_QUERY
 #endif
 
 ! Helper macros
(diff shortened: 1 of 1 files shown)


app/native-multi-image.F90
```diff
@@ -58,7 +58,7 @@ program native_multi_image
 #endif
 
 #ifndef HAVE_COARRAY
-#define HAVE_COARRAY 0
+#define HAVE_COARRAY 1
```
bonachea, line 61: Please revert this change to the default, which would break existing CI jobs for versions predating coarray support.

app/native-multi-image.F90
```diff
@@ -190,6 +207,37 @@ program native_multi_image
...
+#     if HAVE_TEAM
+    write(*,'(A,I3)') "image_index(sca_int_1, [1], get_team()) = ", IMAGE_INDEX(sca_int_1, [1], GET_TEAM())
+    write(*,'(A,I3)') "image_index(sca_int_1, [1], team_number=-1) = ", IMAGE_INDEX(sca_int_1, [1], TEAM_NUMBER=-1)
+#     endif
+#   endif
+#   if HAVE_THIS_IMAGE_COARRAY
+    call status("Testing THIS_IMAGE(coarray)...")
+    write(*,'(A,I3)')  "this_image(sca_int_1) = ", THIS_IMAGE(sca_int_1)
+    write(*,'(A,3I3)') "this_image(sca_int_3) = ", THIS_IMAGE(sca_int_3)
+    write(*,'(A,I3)')  "this_image(sca_int_3, dim=2) = ", THIS_IMAGE(sca_int_3, dim=2)
```
bonachea, line 244: These three lines are the only new output that will be image-specific.

I suggest that all the new write statements but these be conditional on `this_image()==1`, so we don't get flooded with a bunch of duplicate output.

app/native-multi-image.F90
```diff
@@ -190,6 +207,37 @@ program native_multi_image
...
+#   if HAVE_COSHAPE
+    call status("Testing COSHAPE...")
+    write(*,'(A,3I3)') "coshape(sca_int_3) = ", COSHAPE(sca_int_3)
+    write(*,'(A,3I3)') "coshape(sca_int_3, kind=8) = ", COSHAPE(sca_int_3, kind=8)
+#   endif
+#   if HAVE_IMAGE_INDEX
+    call status("Testing IMAGE_INDEX...")
+    write(*,'(A,I3)') "image_index(sca_int_1, [1]) = ", IMAGE_INDEX(sca_int_1, [1])
+#     if HAVE_TEAM
+    write(*,'(A,I3)') "image_index(sca_int_1, [1], get_team()) = ", IMAGE_INDEX(sca_int_1, [1], GET_TEAM())
```
bonachea: You reported in Slack this line appears to be crashing. I've seen some flang codegen bugs with `TEAM_TYPE` that have not been fully diagnosed and remain unresolved.

Until that's resolved, I'd suggest trying a temporary variable here


```suggestion
    res = GET_TEAM()
    write(*,'(A,I3)') "image_index(sca_int_1, [1], get_team()) = ", IMAGE_INDEX(sca_int_1, [1], res)
```

bonachea (commented):
Some initial thoughts..

bonachea:
superseded by #340

PR #341, opened 2026-07-08
Add assertions along all PRIF entry points that prif_init has been called

bonachea:
Many PRIF entry points begin with a call to internal helpers `team_check` or `coarray_handle_check`, so new assertions there cover a large fraction of the public entry points.

The rest are covered by strategic insertion of assertions near the entry point.

`prif_init_called_previously` is promoted to a module-level internal symbol to enable these assertions, and a new variable local to `prif_init` is used to detect potential discrepancies from memory corruption.


ktras (approved):
LGTM

PR #347, opened 2026-07-18
native-multi-image: Add event and putget coverage

bonachea:
Expand native-multi-image with coverage for events and intrinsic scalar put/get.
Adjust CI settings accordingly.

Also enable import of LFortran's iso_fortran_env constants


ktras (approved):
LGTM.

PR #349, opened 2026-07-21
Poison allocated memory when ASSERTIONS are enabled

bonachea:
Memory allocated by `prif_allocate[_coarray]` is uninitialized, but often happens to be zero which can hide problems. When assertions are enabled, poison some of the first page with non-zero values that are more likely to trigger failures in downstream use of uninitialized memory


ktras (approved):
LGTM

PR #359, opened 2026-08-13
Fix #358: install.sh failure with FC in a directory name matching `flang-[1-9][0-9]`

### LINKED ISSUE #358: install.sh failure with FC in a directory name matching `flang-[1-9][0-9]`
Attempting to install with `FC` set to a compiler that resides in a pathname where any directory component contains a substring matching `flang-[1-9][0-9]` currently leads to incorrect behavior.

The failure usually manifests as a `No such file or directory` error on the first `$FPM_FC` invocation within `install.sh`, eg:
```
./install.sh: line 518: /global/cfs/cdirs/m2878/flang/flang-new: No such file or directory
```

This is caused by an overly broad regex in install.sh.

bonachea:
Fixes #358

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -367,7 +367,8 @@ echo "PKG_CONFIG_PATH=$PKG_CONFIG_PATH"
 
 FPM_FC="$($REALPATH $(command -v $FC))"
 if [[ $FPM_FC == *flang* ]]; then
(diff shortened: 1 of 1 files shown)


ktras (approved):
LGTM

PR #367, opened 2026-09-09
Install.sh overhaul

### LINKED ISSUE #241: Add explicit control of assertions
The Caffeine library uses the Assert library to add many correctness checking assertions. This includes both internal sanity checks and also validation of input to PRIF calls. These assertions are extremely valuable for correctness validation and debugging, but also incur a runtime performance overhead.

As such, it's important to ensure clients of Caffeine choose the appropriate build mode for their use case, namely:

* Assertions ON: For correctness testing and validation
* Assertions OFF: For performance testing and production use

Currently there's no consistent explicit control for this build mode:

* `install.sh` currently defaults to Assertions ON, unless it sees a `-[DU]ASSERTIONS` argument in `CPPFLAGS` or `FFLAGS` (which then take precedence). This can also be overridden at `fpm` invocation time.
* Builds without `install.sh` that pass no `-[DU]ASSERTIONS` argument will result in the default behavior of Assertions OFF

To make matters worse, assertions are currently controlled in caffeine.c using a completely separate mechanism: they are enabled by default and `-DNDEBUG` in the C compiler flags turns them off.

This is messy and error-prone.

We should deploy a robust and uniform mechanism to control the Caffiene build mode.

### LINKED ISSUE #368: install.sh should warn or fail early for unsupported Fortran compiler versions
Hey! 👋 

Thank you for a very cool project! 

I tried setting this up for the first time. I initially hit some issues with gfortran. On Pop!_OS 22.04, install.sh accepted GNU Fortran 11.4 and proceeded to build GASNet-EX. The subsequent Caffeine compilation then failed in prif_private_s.F90 with:

Error: Character argument ‘str’ must be length 1 because procedure
‘caf_fatal_error’ is BIND(C)

For compilation I used

```
 env FC=gfortran CC=gcc CXX=g++  ./install.sh --network=smp
```
On my system this resolved to GCC/gfortran 11.4. Rebuilding explicitly with GCC/gfortran 13 resolved the problem, and the tests passed.

The README lists gfortran 13–15 as the regularly tested versions. Would it make sense for the installer to detect the selected compiler version before building dependencies and either:

 - reject versions known to be incompatible; or
 - warn when the version is outside the tested range?

This would provide a more actionable error and avoid building GASNet-EX before discovering a compiler incompatibility.

bonachea:
Performs some overdue cleanup on the `./install.sh` script to improve robustness and maintainability, in preparation for future feature expansion. 

Highlights:

1. Adds `--{enable,disable}-debug` option, mirroring the GASNet option, to control production vs debugging mode
2. Converts the default install behavior to **disable assertions**. Use `--enable-debug` to restore assertion checking.
3. Clean up and expand the Caffeine pkg-config file output (i.e. `caffeine*.pc` files)
4. Updates bash-isms throughout to be more concise and maintainable
5. CI updates for install.sh invocation and coverage
6. Updates documentation

See individual commit messages for more details.

Fixes #241
Fixes #368


README.md
```diff
@@ -48,29 +48,34 @@ their solution to support Fortran's multi-image parallel features.
...
-* [make]
-* [git]
-* [curl]
-
-The script will invoke these if present in a user's `PATH`.
-If not present, the script will ask permission to use [Homebrew] to install the relevant package
+
+The `install.sh` script uses the following packages to build Caffeine:
+* Fortran and C compilers. We [regularly test](https://github.com/BerkeleyLab/caffeine/actions?query=branch%3Amain) with:
+   - LLVM Flang versions 19:23,
```
rouson:
```suggestion
   - LLVM Flang versions 19 or newer,
```

README.md
```diff
@@ -48,29 +48,34 @@ their solution to support Fortran's multi-image parallel features.
...
+   - GNU Fortran versions 13:16, and
```
rouson:
```suggestion
   - GNU Fortran versions 13 or newer, and
```

README.md
```diff
@@ -48,29 +48,34 @@ their solution to support Fortran's multi-image parallel features.
...
+   - LFortran versions 0.64:
```
rouson:
```suggestion
   - LFortran versions 0.64 or newer
```

install.sh
```diff
@@ -20,9 +21,10 @@ USAGE:
                       ofi: OpenFabrics Interfaces
                       ucx: Unified Communication X
  --prereqs          Display a list of prerequisite software.
-                    Default prefix='\$HOME/.local/bin'
  --verbose          Show verbose build commands
- --yes              Assume (yes) to all prompts for non-interactive build
+ --yes              Assume (yes) to all prompts for non-interactive install
```
rouson:
```suggestion
 --yes              Assume "yes" to all prompts for non-interactive install
```

rouson (approved):
LGTM

PR #371, opened 2026-09-15
Update Assert and Julienne dependencies

bonachea:
This PR updates dependencies to the newly released versions of Assert and Julienne. 

Together these should provide more robust integration for use cases outside our normal CI workflows (e.g. building a Julienne-enabled testsuite  for a scientific library using flang `-fcoarray` and linking Caffeine and possibly a fourth-party assertions package).


rouson (approved):
LGTM

